Issue #20, opened 2019-08-15
Chapter 1 future improvements


ttimbers:
- actually use `select` and `filter` when creating the plot (more convincing/motivating) and showing the code from nachos to cheesecake 

Above has been done for filter.

ttimbers:
Reopening as I have some future thoughts:

1. Should we have sections that address everything in the learning objectives? Typically my answer is yes, however we have a lot of scaffolding/explanations in the worksheet for this chapter, and given this is a pre-reading before the first lecture I don't anticipate all people reading this before the worksheet, so the worksheet needs to stand alone. And I worry about the chapter being super redundant with the worksheet if we add a lot more... 

2. How do we link to the worksheets and tutorials? I am thinking of putting links to them at the end of each chapter? Also, with the link, a reminder to UBC students that they should be accessing their worksheets & tutorials via Canvas...

trevorcampbell:
@ttimbers I think this is all handled now in v0.5.0?

ttimbers:
Now we instead use `mutate` in nachos to cheesecake, but I think its fine.

PR #21, opened 2019-08-15
Chapter1

ttimbers:
Ok @trevorcampbell I have tried to improve Chapter 1. I still need to upgrade the video audio and address #20 but I think this could be good enough to go to production if needed. So ready for your feedback.


trevorcampbell:
Some things I changed in my most recent commit after review:

- reorganized some paragraph text for style
- fixed spelling
- added [blue text links](fake_link)
- fixed broken worksheet link
- added *emph*s for newly defined words
- added a little blurb about how Jupyter does syntax highlighting for us
- suppressed us_data assignment read_csv output so that the following text makes more sense
- changed "subsetting" to "creating"/"obtaining a subset" (subset not a verb)
- fixed buggy numbered list of columns (added newline before list)
- added line breaks in long paragraphs for broader editor compatibility (eg vim)
- renamed "creating visualizations in R" to "exploring data with visualizations" (seems to fit better with the other titles, less overlap with title /concept of Ch 4)
- added a blurb about using print() to visualization section
- added extra info greyed-out blurb about multi-line ggplot expressions

Feel free to remove blurbs I added if you think they're too much, or to reword things as you see fit!

Things left to do:
- there's something wrong with state_property_vote vs state_property_data. In the repo, the filename is "state_property_data.csv", but the R code runs on "state_property_vote.csv" and I can't seem to find that file...I might just be being brain dead...
- in the political plot, democrats should be blue and GOP should be red ( party colours! :) )
- for the first scatter interpretation, this looks more to me like an outlier
  - later on it's basically shown to be an outlier. Maybe worth briefly mentioning outliers vs nonlinearities
- there's still the mention of web scraping in Ch 2 at the end of Ch 1, which I imagine should be removed?

trevorcampbell:
> Ok @trevorcampbell I have tried to improve Chapter 1. I still need to upgrade the video audio and address #20 but I think this could be good enough to go to production if needed. So ready for your feedback.

We can definitely leave the video/audio for later on! :)

ttimbers:
> there's something wrong with state_property_vote vs state_property_data. In the repo, the filename is "state_property_data.csv", but the R code runs on "state_property_vote.csv" and I can't seem to find that file...I might just be being brain dead...

In the chapter1 branch, I have a file called `state_property_vote.csv` that lives in the root. I put it into the root because we haven't explained paths yet. So maybe that it why you didn't see it. But its there when I check out the branch.

> in the political plot, democrats should be blue and GOP should be red ( party colours! .... ok this is super unimportant and probably not worth doing 😆 )

Ohhh! I agree! Done!

> for the first scatter interpretation, this looks more to me like an outlier
later on it's basically shown to be an outlier. Maybe worth briefly mentioning outliers vs nonlinearities

I don't disagree, added a sentence to muse/caution about this.

> there is the greyed out "for advanced readers" blocks. I couldn't think of a nice consistent heading to give them... they aren't really only for people who have used R before, but I don't want to go as dry as "additional info".

For now I have renamed this: "In case you want to know more (optional)". If we come up with a better header later we can switch it.

PR #23, opened 2019-09-08
Chapter2

ttimbers:
~~I still would like to improve the following:~~

~~1. better explain the purpose of `dbplyr::collect` (i.e., I am not convinced it's needed for a lot of downstream analysis)~~
~~2. Add a section on connecting to PostrgreSQL~~ fixes added to address these


ttimbers:
@trevorcampbell - I have done what I think is a sufficient draft of chapter 2 to ship to the students. Please read it over and let me know if there are any necessary fixes before we merge this.

trevorcampbell:
Read through it this morning. I made some minor changes to wording (nothing major) as well as:

- `dbListTables` is from the `DBI` package, not `RSQLite`; changed
- changed chapter learning objectives to use `-` instead of `*` as list delimiter (doesn't change the output format, but fixes some bugs with italics). I also made an issue to go through the whole book later on and make this change for lists.
- removed sub-sub-sub-sub section for absolute/relative path examples (replaced with bold text)
- I found the description of relative/absolute to be a bit confusing, so I just added an example before the video
- added an example of `nrow` failing on `tbl` objects

I'm OK with merging as-is, but if you want to take a light pass please do! Feel free to merge whenever you want though.

Things I'll make issues for for future:
- changing `*` to `-` for lists to avoid markdown problems
- more explanation of why we should care about databases (can take from worksheet 02 / tut 02)
- add URL links to the different database types in 2.6 when introduced

ttimbers:
Wonderful @trevorcampbell - I appreciate your feedback and fixes. Merging and shipping now!

Issue #19, opened 2019-08-06
Style consistency


trevorcampbell:
Once things are settling we should go through and do a pass to make sure style is consistent. Eventually it's probably worth having a section in the README about this. A list of potential things to check below (will be updated over time):

- section heading style (caps, no caps)
- whether data science should be capitalized 
  - no, since other fields e.g. chemistry, statistics, computer science, etc are not capitalized)
- when to use numerals vs words for numbers 
  - "four common problems" vs "4 common problems", "10km race" vs "ten km race"
  - for #s with units, use numerals; for #s in text without units, use words (so four common pbms, 10km race)
- data set vs dataset (I typically use dataset)
  - but we do "data frame", which I guess I agree with...so let's do "data [space] *" in general
- oxford comma (yes!)
- I don't believe "subset" is a verb -- should be "obtain/create/extract a subset"
- functions should be typeset `like_this` without parentheses
- use `-` for lists, not `*`
- use `"` for strings, not `'`
- `summarize`, not `summarise` (Canadian spelling)
- each ```{r}``` block should have a name of the form `##-name-of-block` with ## = chapter number and words separated by hyphens
- center each image (if appropriate) using `<center>  [image code] </center>`
- consistent form of "K-nearest neighbours" and abbrev "KNN" (different names/abbrevs currently throughout)

trevorcampbell:
added to style guide

PR #26, opened 2019-09-16
Chapter3

ttimbers:
Done draft of chapter 3 edits for this year. @trevorcampbell - please review.


trevorcampbell:
I've made lots of minor edits for style/grammar. Below are comments for you @ttimbers to handle. The action items (denoted by checkboxes) I left for you shouldn't require much work, but I figured they were large enough that I didn't just want to go in and change myself. There are a few larger things I added / changed that are also described below; feel free to edit.

- [ ] In 3.4.2, "there are many tools to tidy your data" isn't a reason that tidy data is important. The 2nd and 3rd reasons are good though. I would just move this text later in the paragraph to say that there are many tools available for this and we'll cover them now.

- [ ] In 3.4.3, I wouldn't necessarily call this data untidy. It really depends on what you consider your "observation" to be. For example, if your observation is "the results for 1 year", then it's tidy. If the observation is "the outcome of a single candidate's run for office" then it's untidy. E.g. if I wanted to do something contrived like "2D scatter plot of the # of "e"s in the winner and runner up's names for each year", then I'd want the original data format. We could add a remark about "tidiness" being a matter of what you consider your observations and variables to be.

- [ ] I found the introduction of `gather` to be super confusing. Also, right after "to accomplish this data transformation we will use the tidyverse function gather", we should have a brief explanation of what gather does intuitively. I also tried to improve the wording of the things, but it isn't much better honestly.

- changed "when your code line exceeds 80 chars" to "when your code line gets too long to read clearly", since 80ch isn't a hard limit and might confuse students

- throughout, tidy is specified as (1) each row is a single obs, (2) each variable is a single col, (3) each value... . The second item in the list is in opposite order of the other two: should be "each column is a single variable". I've changed this already as part of my minor edits but figured I'd mention it in case there was a reason to have the inconsistent order.

- [ ] the US presidential election result figure should probably just have one of the bar groups, since the information is just duplicated between the winner + runnerup groups (US de facto only has two parties)

- [ ] I changed the intro to pipes to focus on *why* they matter more. Take a look and feel free to edit. One thing you might want to change is to get rid of the pipe example at the beginning (it has 3 lines, which is something you cover a few paragraphs down). I will let you make that call; the nice thing about the example there is that it shows immediately how much clearer piped code is. You could also move the greyed out note about multi-line pipe code a few paras down to here. All up to you.

- on every code block, we should provide an identifier of the form ##-name-with-hyphens for consistency (we can do this later, I added to the style consistency issue #19 )

- I'm not sure the pipe `%>%` actually increases efficiency of anything. I don't think it stops temporaries from being created; it just stops you from having to name them. I've refocused wording on readability.

- for future editing, "subset" isn't a verb; I go with "obtain / extract / create a subset". I changed these in this chapter.

- I changed `summarise` to `summarize` (canadian spelling) and added a note to the style consistency issue #19 

- [ ] I didn't touch the text starting with "What if you need to add other arguments...". In principle I like the idea of introducing anonymous functions, but I think it gets swept under the rug a bit and is quite an advanced concept to have this early. Here is how I'd handle this:
    - remove the anon functions; use mutate w/ pipe to set the NAs to -infty (or 0 here, which I think is fine for this data) before maxing. Make anon functions an optional section with a proper ## heading.

ttimbers:
> In 3.4.3, I wouldn't necessarily call this data untidy. It really depends on what you consider your "observation" to be. For example, if your observation is "the results for 1 year", then it's tidy. If the observation is "the outcome of a single candidate's run for office" then it's untidy. E.g. if I wanted to do something contrived like "2D scatter plot of the # of "e"s in the winner and runner up's names for each year", then I'd want the original data format. We could add a remark about "tidiness" being a matter of what you consider your observations and variables to be.

I disagree for this case. When the table is like this:

```
## # A tibble: 10 x 3
##    election_year winner        runnerup       
##            <dbl> <chr>         <chr>          
##  1          2016 Donald Trump  Hillary Clinton
##  2          2012 Barack Obama  Mitt Romney    
##  3          2008 Barack Obama  John McCain    
##  4          2004 George Bush   John Kerry     
##  5          2000 George Bush   Al Gore        
##  6          1996 Bill Clinton  Bob Dole       
##  7          1992 Bill Clinton  George Bush    
##  8          1988 George Bush   Michael Dukakis
##  9          1984 Ronald Reagan Walter Mondale 
## 10          1980 Ronald Reagan Jimmy Carter
```

What is Y in the context of your question "the results for 1 year"? The Y (response) variable should be in one column if the data is tidy.

trevorcampbell:
> What is Y in the context of your question "the results for 1 year"? The Y (response) variable should be in one column if the data is tidy.

I'll change the situation a little bit to (hopefully) make the example clearer. Suppose instead my data were

```
## # A tibble: 10 x 5
##    election_year winner  winner_lean      runnerup    runnerup_lean    
##            <dbl> <chr>    <dbl>      <chr>          <dbl>    
##  1          2016 Donald Trump   -0.3   Hillary Clinton   0.5
##  2          2012 Barack Obama    0.6   Mitt Romney      -0.1
##  3          2008 Barack Obama    0.6   John McCain      -0.2
##  4          2004 George Bush    -0.2   John Kerry        0.5
##  5          2000 George Bush    -0.2   Al Gore           0.8     
...
```
and I wanted to do a scatter of winner vs runnerup political leaning measured on [-1, 1], and label each point on the scatter with the two names (winner + runnerup) with the winner's name bolded. Then the above data would be considered "tidy", while using the equivalent to the second version in the text (with "result" in a separate variable) would be a pain.

This is because here the "observation"/"row" is total outcome and info for 1 year's election, while the second version of the data considers an observation to be a single candidate's run.

> I do however agree that '"tidiness" being a matter of what you consider your observations and variables to be', but it's tricky here because they don't really know what random variables are yet here... I will add a vagary that cautions about this.

I don't think we need to talk about random variables -- I was using "variable" in the parlance of the textbook so far, i.e., "column".

ttimbers:
I really like the intro you added to the pipes section!

ttimbers:
> I didn't touch the text starting with "What if you need to add other arguments...". In principle I like the idea of introducing anonymous functions, but I think it gets swept under the rug a bit and is quite an advanced concept to have this early. Here is how I'd handle this:

> remove the anon functions; use mutate w/ pipe to set the NAs to -infty (or 0 here, which I think is fine for this data) before maxing. Make anon functions an optional section with a proper ## heading.

They need to use anonymous functions with map later in the course to map other functions with multiple arguments so we are kind of stuck introducing it here... Yes, this isn't the most motivating example, but its a simple demo before they have to start using it in modelling (used in clustering to do kmeans clustering for a range of cluster sizes when choosing k). I think it fits here because `group_by` + `summarize` is an abstraction of iteration, and so is `map`. We need to give them these tools to iterate given that we are not teaching for loops (and these are arguably also the strategies anyone should use to iterate in R when working with vectorized objects).

ttimbers:
Thanks for the review, changes and feedback. Merging!

PR #41, opened 2020-01-30
scaling fix for chapter 6

ttimbers:


trevorcampbell:
Looks good! But I see this right after *More than two explanatory variables*:

(I don't think you changed this in the edit but maybe worth fixing while we're here anyway -- I tried a few things to suppress but `echo = FALSE` and `warning = FALSE` didn't work)

## 
## Attaching package: 'plotly'
## The following object is masked from 'package:ggplot2':
## 
##     last_plot
## The following object is masked from 'package:stats':
## 
##     filter
## The following object is masked from 'package:graphics':
## 
##     layout

ttimbers:
Ah, I need to add `message=FALSE` in that code chunk. Probably we should just set this for all chapters as a global option (for all chapters) as this book almost never wants such messages print. I'll just fix it in the chunk here, and open an issue that we need to do this for all chapters.

ttimbers:
@trevorcampbell - ready for you to merge I think

PR #42, opened 2020-02-01
replaced scale with prePRocess in all chapters; fixed over/underfit e…

trevorcampbell:
- replaced `scale` with `preProcess` in all chapters
- made sure to do test/train *and* X/Y splits before preProc
- fixed broken under/overfit example in Ch 7


ttimbers:
@trevorcampbell - what is going on with this PR?

trevorcampbell:
> @trevorcampbell - what is going on with this PR?

Looks like it is still pending your review! If you agree with the changes (which were long enough ago now that I don't really recall, but the notes in the original message should guide) then I'd merge it into `dev` rather than `master`

trevorcampbell:
(there are a few open issues about this, e.g. #43  )

trevorcampbell:
I believe this PR is superceded by #55 on Tidymodels, so I'm closing it.

PR #54, opened 2020-07-09
Update 02-wrangling.Rmd

leem44:
removing comma from  so image will display in section 3.4.1


leem44:
@ttimbers very minor fix to the textbook, but thought I should just PR in case!

trevorcampbell:
Good catch. Thanks for making a PR -- now I'm aware of the issue and I will make the same change in the `dev` branch :)

PR #55, opened 2020-07-16
Tidymodels

trevorcampbell:
This PR is for updating chapters 6, 7, 8, and 9 to the `tidymodels` framework.

Also added the `libglpk-dev` package and `kknn` R library to the Dockerfile.

The actual prose in these chapters could definitely still use some work, but I'd like to (a) get feedback on what's there and (b) merge the big code changes into `dev` before doing edit round 2.


leem44:
Overall I like the changes you made with respect to tidymodels! It had a good flow and was well written!

Here are my comments: 
Chapter 6: 
- Learning objectives "Combine preprocessing and model training using a `workflow`" -- perhaps we should specify as "tidymodels workflow" or something along those lines
- Sec 6.6
  - would be helpful to give more explanation of the "weight_func"
   - would also be helpful to give some quick interpretation of the output of knn_fit 
- Sec 6.7.1 
  - This is just a general comment about this section: I wonder if it would be useful to give a formula here or even just show step by step how we subtract the mean and then divide by the sd in the dataset. I find students confused with what we are doing exactly when we are centering, scaling
  - It says "We now use the prep function to create an object that represents..." but I think the `prep` function was used in the code chunk above
  - I like the picture of the standardized vs. nonstandardized data now with the lines! It looks good and illustrates the point better than before
- sec 6.8 
  - would be helpful to include a short explanation of the output of knn_fit
  - I see "NN`{r 06-setup, include=FALSE} knitr::opts_chunk$set(message = FALSE)" below the first graph when I render the book and then the text below is in a code cell. 
- sec 6.10 feels like it should come earlier or be integrated into the earlier sections
- Overall I think it would be useful to have an overview of the `tidymodels` framework/workflow at the start of section 6.6 e.g. discussing the recipe, the "bake" step, workflow etc. I initially found all the different pieces of the "recipe" in tidymodels a tad confusing so an overview of the different pieces might help

Chapter 8 
- sec. 8.4 -- looks like we're scaling the predictors here in the simple linear regression, I would argue that we shouldn't perform scaling here because I think it affects the interpretability and then there's also a mismatch between the output and graphs etc

trevorcampbell:
Handled all except explaining scaling/centering -- making an issue for that for later!

PR #60, opened 2020-10-25
Ch 11 - Inference

leem44:
@ttimbers @trevorcampbell -- very rough draft of the intro to inference chapter! Please have a look and let me know anything you think should be changed!


trevorcampbell:
It looks like this PR is based on an old version of `master`. Normally this wouldn't matter, but `bookdown::render_book` runs into an exception:

```
label: 05-upsample-cancer
Quitting from lines 2651-2657 (_main.Rmd) 
Error in attributes(out) <- attributes(col) : 
  'names' attribute [360] must be the same length as the vector [357]
```

Can you rebase the branch on most current `master`?

trevorcampbell:
Comments:

- rather than sending detailed comments, I went through made some changes directly; take a look at these
- are all images licensed in a way that lets us use them?
- the notes about asymptotics / normality / etc are a bit too technical, no? I haven't taught the inference weeks before so not sure how much detail we go into there. Like, I'm not sure most students even know what a normal distribution is (...or a distribution, for that matter....)
- added a few [TODO] notes to the bootstrap subsection
- I don't know if I like telling the students that they're computing "confidence intervals". The bootstrap distribution percentiles are not necessarily confidence intervals (for the same reason the sampling distribution intervals are not necessarily confidence intervals). In *practice*, they tend to be reasonable estimates of proper confidence intervals. But that's a nuance I'm not sure I want to try to impart in a 100-level course. Maybe say "interval estimate" and then refer the students to more advanced material on "confidence intervals"???
- compile warnings:

```
[WARNING] Duplicate identifier 'my-section' at line 2282 column 1
[WARNING] Duplicate identifier 'my-section' at line 2312 column 1
[WARNING] Duplicate identifier 'my-section' at line 2695 column 1
[WARNING] Duplicate identifier 'my-section' at line 2741 column 1
```
make sure each section follows naming convention from previous chapters

-

leem44:
@trevorcampbell Thank you for the edits!!!

> Overall, awesome. Comments:
> 
> * rather than sending detailed comments, I went through made some changes directly; take a look at these
> * are all images licensed in a way that lets us use them?
- yes the one Timbits image I made sure was CC0 and the other images I made so I think it should be okay!
> * the notes about asymptotics / normality / etc are a bit too technical, no? I haven't taught the inference weeks before so not sure how much detail we go into there. Like, I'm not sure most students even know what a normal distribution is (...or a distribution, for that matter....)
- I removed the "Notes" altogether since its too technical
- I also use the word "distribution" quite a bit in the writing... I figured it should be fine to just use the word "distribution" but let me know if you think I should mention more explicitly what it means somewhere
> * added a few [TODO] notes to the bootstrap subsection
- done! 
> * I don't know if I like telling the students that they're computing "confidence intervals". The bootstrap distribution percentiles are not necessarily confidence intervals (for the same reason the sampling distribution intervals are not necessarily confidence intervals). In _practice_, for reasonable statistics, they tend to be reasonable estimates of proper confidence intervals (essentially because asymptotic results hold). But that's a nuance I'm not sure I want to try to impart in a 100-level course. Maybe say "interval estimate" and then refer the students to more advanced material on "confidence intervals"???
- I changed this to phrase to "approximate bootstrap percentile CI" -- we can change this as we see fit...

> * compile warnings:
> 
> ```
> [WARNING] Duplicate identifier 'my-section' at line 2282 column 1
> [WARNING] Duplicate identifier 'my-section' at line 2312 column 1
> [WARNING] Duplicate identifier 'my-section' at line 2695 column 1
> [WARNING] Duplicate identifier 'my-section' at line 2741 column 1
> ```
> 
> make sure each section follows naming convention from previous chapters
- I will look into this warning 

- also changed "color" to "flavour" in creation of population of timbits 
- hid the code creating the population 

@ttimbers ready for you to take a look whenever you get a chance!

10-inference.Rmd
```diff
@@ -0,0 +1,806 @@
...
+  ggp_data <- ggplot_build(dist)
+  min(ggp_data$data[[1]]$xmin)
+}
+```
+
+
+## Overview
+In almost all data analysis tasks in practice, we want to draw conclusions about some unknown
+aspect of a population of interest based on observed data sampled from that
+population; we typically do not get data on the *full* population.
```
ttimbers, line 25: I would say this is true mainly for inferential questions. In predictive questions, we are interested mainly in predicting future values for individuals, or what features are important for predicting future values for individuals. Maybe soften this statement to "Another common data analysis task in practice is to draw conclusions...

```suggestion
Another common data analysis task in practice is to draw conclusions about some unknown
aspect of a population of interest based on observed data sampled from that
population; we typically do not get data on the *entire* population.
```

10-inference.Rmd
```diff
@@ -0,0 +1,806 @@
...
+## Why do we need sampling? 
+Statistical inference can help us decide how quantities we observe in
+a subset of data relate to the same quantities in the broader
+population. Here is an example question that we might use statistical inference to answer:
+
+*What proportion of all undergraduate students in North America own an iPhone?*
+
+In the above question, we are interested in making a conclusion about *all*
+undergraduate students in North America.  This is our **population**: 
+in general, the population is the complete collection of individuals or cases we are interested in studying.
```
ttimbers, line 56:
```suggestion
In the above question, we are interested in making a conclusion about *all*
undergraduate students in North America; this is our **population**.
In general, the population is the complete collection of individuals or cases we are interested in studying. 
```

10-inference.Rmd
```diff
@@ -0,0 +1,806 @@
...
+
+In the above question, we are interested in making a conclusion about *all*
+undergraduate students in North America.  This is our **population**: 
+in general, the population is the complete collection of individuals or cases we are interested in studying. 
+Further, in the above question, we are interested in computing a quantity&mdash;the proportion
+of iPhone owners&mdash;based on the entire population. This is our **population parameter**:
+in general, a population parameter is a numerical characteristic
+of the entire population. In order to compute this number in the example above, we would need to ask 
+every single undergraduate in North America whether or not they own an iPhone. In practice,
+directly computing population parameters is often time-consuming and costly, and sometimes impossible.
```
ttimbers, line 62:
```suggestion
Further, in the above question, we are interested in computing a quantity&mdash;the proportion
of iPhone owners&mdash;based on the entire population. This is our **population parameter**.
In general, a population parameter is a numerical characteristic
of the entire population. In order to compute this number in the example above, we would need to ask 
every single undergraduate in North America whether or not they own an iPhone. In practice,
directly computing population parameters is often time-consuming and costly, and sometimes impossible. 
```

10-inference.Rmd
```diff
@@ -0,0 +1,806 @@
...
+of the entire population. In order to compute this number in the example above, we would need to ask 
+every single undergraduate in North America whether or not they own an iPhone. In practice,
+directly computing population parameters is often time-consuming and costly, and sometimes impossible. 
+
+A more practical approach would be to collect measurements for a **sample**: a subset of
+individuals collected from the population. We can then compute a **sample statistic**&mdash;a numerical
+characteristic of the sample&mdash;that estimates the population parameter. For example, if we 
+randomly selected 100 undergraduate students across North America (the sample) and computed the fraction of those
+students who own an iPhone (the sample statistic), we might suspect that that fraction is a reasonable
+estimate of the full population fraction.
```
ttimbers, line 69: We should keep the language consistent here, stick with proportion:

```suggestion
A more practical approach would be to collect measurements for a **sample**: a subset of
individuals collected from the population. We can then compute a **sample statistic**&mdash;a numerical
characteristic of the sample&mdash;that estimates the population parameter. For example, if we 
randomly selected 100 undergraduate students across North America (the sample) and computed the proportion of those
students who own an iPhone (the sample statistic), we might suspect that that proportion is a reasonable
estimate of the proportion of students who own an iPhone in the entire population. 
```

ttimbers: Also, do we want to use point estimate in the overview and then switch to sample statistic here?

10-inference.Rmd
```diff
@@ -0,0 +1,806 @@
...
+individuals collected from the population. We can then compute a **sample statistic**&mdash;a numerical
+characteristic of the sample&mdash;that estimates the population parameter. For example, if we 
+randomly selected 100 undergraduate students across North America (the sample) and computed the fraction of those
+students who own an iPhone (the sample statistic), we might suspect that that fraction is a reasonable
+estimate of the full population fraction. 
+
+<center>
+![Figure 11.1: Population versus sample](img/population_vs_sample.svg){}
+</center>
+
```
ttimbers, line 74: I think we need to add some motivation to our iPhone question. Who would be interested in asking and answering this question? And why? Maybe a cell phone retailer that sells iPhones and wants to understand the amount of available market left? Or maybe they sell iPhone accessories, or are thinking of starting to, and they want to estimate how big the market might be?

10-inference.Rmd
```diff
@@ -0,0 +1,806 @@
...
+## Sampling distributions
+
+### Sampling distributions for proportions
+
+Let's start with an illustrative (and tasty!) example. Timbits are 
+bite-sized doughnuts sold at Tim Hortons, a popular Canadian-based fast-food restaurant
+chain founded in Hamilton, Ontario, Canada. 
+
+![Timbits. Source: wikimedia.org](https://upload.wikimedia.org/wikipedia/commons/thumb/7/75/Timbits2.jpg/1600px-Timbits2.jpg){width=50%}
```
ttimbers, line 101: I think we can create our own timbits photo 😉  I volunteer for this job 😋 !

10-inference.Rmd
```diff
@@ -0,0 +1,806 @@
...
+estimate of the full population fraction. 
+
+<center>
+![Figure 11.1: Population versus sample](img/population_vs_sample.svg){}
+</center>
+
+Note that proportions are not the *only* kind of population parameter we might be interested in.
+Let's consider another example question that we might tackle with statistical inference:
+
+*What is the average price-per-night of one-bedroom apartment rentals in Vancouver, Canada?*
```
ttimbers, line 78: Same for this question. Let's add a sentence or two motivating this question. Perhaps undergraduates starting at UBC are looking for an apartment and they want to know what is an "average" price? If we do go with undergraduates to motivate this, we might have to use a more realistic rental unit, studio apartment? bedroom in shared accommodation?

10-inference.Rmd
```diff
@@ -0,0 +1,806 @@
...
+
+Here, the population consists of all one-bedroom apartment rental offerings in Vancouver, and the population
+parameter is the *average price-per-night*. But even within this one example, we could also be interested
+in many other population parameters: the median price, the fraction of 
+one-bedroom apartments that cost more than \$200 per night,
+the standard deviation of the price, and the list goes on.
+If we were somehow able to observe the whole population of one-bedroom apartments in Vancouver, 
+we could compute each of these numbers exactly; therefore these are all population parameters. 
+There are many kinds of observations and population parameters that you will run into in practice,
+but in this chapter we will focus on two settings:
```
ttimbers, line 88: I think we should emphasize that you choose the parameter to estimate based on the question you are asking. For example, if you are interested in the most commonly observed values in the population, then the mean generally, is a good parameter to estimate if your data is quantitative. If however, you were interested in how much values vary in the population, then you would want to estimate something else, like the standard deviation or variance. Or if you were interested in knowing the highest values in a population, you would estimate a quantile (e.g., a 95th percentile).

ttimbers (changes requested):
@leem44 @trevorcampbell - here are some suggestions and comments for the lead into this chapter. I'm off to teach, but after teaching I will comment on the rest of the chapter.

10-inference.Rmd
ttimbers: Maybe the scenario is that they are trying to make a budget?

10-inference.Rmd
```diff
@@ -0,0 +1,806 @@
...
+samples to vary? In order to understand this, we will simulate more samples
+of size 40 from our population of Timbits, and calculate the 
+proportion of chocolate Timbits in each sample. We can then
+construct the distribution of sample proportions we calculate. The distribution
+of the statistic for all possible samples of size $n$ from a population is
+called a **sampling distribution**. The sampling distribution will help us see
+how much we would expect our sample proportions from this population to vary
+for samples of size 40. Below we again use the `rep_sample_n` to take samples
+of size 40 from our population of Timbits, but we set the `reps` argument
+to specify the number of samples to take.
```
ttimbers, line 179: We have not yet defined n
```suggestion
of the estimate for all possible samples of a given size (which we commonly refer to as $n$) from a population is
called a **sampling distribution**. The sampling distribution will help us see
how much we would expect our sample proportions from this population to vary
for samples of size 40. Below we again use the `rep_sample_n` to take samples
of size 40 from our population of Timbits, but we set the `reps` argument
to specify the number of samples to take, here 1000.
```

10-inference.Rmd
```diff
@@ -0,0 +1,806 @@
...
+
+```{r 11-example-proportions5, echo = TRUE, message = FALSE, warning = FALSE}
+samples <- rep_sample_n(virtual_box, size = 40, reps = 1000)
+head(samples)
+tail(samples)
```
ttimbers, line 184: Here the use of `head` and `tail` are warranted so that we can see both ends of the tibble. If we drop `head` from earlier on in the book we might then need to explain what is happening here.

10-inference.Rmd
```diff
@@ -0,0 +1,806 @@
...
+```
+
+Notice the column `replicate` is indicating the replicate with which each
+Timbit belongs. Since we took 1000 samples of size 40, there are 1000 replicates.
```
ttimbers, line 188:
```suggestion
Notice the column `replicate` is indicating the replicate, or sample, with which each
Timbit belongs. Since we took 1000 samples of size 40, there are 1000 replicates.
```

10-inference.Rmd
```diff
@@ -0,0 +1,806 @@
...
+
```
ttimbers, line 189:
```suggestion
Now that we have taken 1000 samples, to create a sampling distribution of sample proportions for samples of size 40, we need to calculate the proportion of chocolate Timbits for each sample, $\hat{p_\text{chocolate}}$:
```

10-inference.Rmd
```diff
@@ -0,0 +1,806 @@
...
+```{r 11-example-proportions6, echo = TRUE, message = FALSE, warning = FALSE}
+sample_estimates <- samples %>% 
+    group_by(replicate) %>% 
+    summarise(sample_proportion = sum(flavour == "chocolate") / 40)
+head(sample_estimates)
+tail(sample_estimates)
+```
+
```
ttimbers, line 197:
```suggestion
Now that we have calculated the proportion of chocolate Timbits for each sample, $\hat{p_\text{chocolate}}$, we can visualize the sampling distribution of sample proportions for samples of size 40:
```

10-inference.Rmd
```diff
@@ -0,0 +1,806 @@
...
+```{r 11-example-proportions7, echo = TRUE, message = FALSE, warning = FALSE,fig.cap = "Sampling distribution of the sample proportion for sample size 40"}
+sampling_distribution <-  ggplot(sample_estimates, aes(x = sample_proportion)) +
+    geom_histogram(fill="#0072B2", color="#e9ecef", binwidth = 0.05) +
```
ttimbers, line 200: Can we use human readable names for the colours here? These hex colours are a bit confusing/distracting for first year students. Also, I think `bins` is easier to understand and use in histogram than `bindwidth` (where you need to think about scale for each kind of data you are working with).

10-inference.Rmd
```diff
@@ -0,0 +1,806 @@
...
+
+A more practical approach would be to collect measurements for a **sample**: a subset of
+individuals collected from the population. We can then compute a **sample statistic**&mdash;a numerical
+characteristic of the sample&mdash;that estimates the population parameter. For example, if we 
+randomly selected 100 undergraduate students across North America (the sample) and computed the fraction of those
+students who own an iPhone (the sample statistic), we might suspect that that fraction is a reasonable
+estimate of the full population fraction. 
+
+<center>
+![Figure 11.1: Population versus sample](img/population_vs_sample.svg){}
```
ttimbers, line 72: We should have a figure caption for all figures (we will need to do this for the timbits image too). I think it's easiest to do this using `knitr::include_graphics`. You might want to include `fig.retina = 2` as a code chunk option to make the image crisp and sharp, and `fig.cap` to specify the caption.

10-inference.Rmd
```diff
@@ -0,0 +1,806 @@
...
+sample proportions range from about `r round(min(sample_estimates$sample_proportion), 1)` to 
+about `r round(max(sample_estimates$sample_proportion), 1)`. In fact, we can calculate
+the mean and standard deviation of the sample proportions. 
+
+```{r 11-example-proportions8, echo = TRUE, message = FALSE, warning = FALSE}
+sample_estimates %>% 
+  summarise(mean = mean(sample_proportion), sd = sd(sample_proportion))
+```
+
+We notice that the sample proportions are centred around the population
```
ttimbers, line 216: We need more oomf here! It's actually pretty cool and incredible that the centre of the sampling distribution is the same as the population parameter. Here we don't actually see that, because we didn't take enough replicates. Let's actually take 10,000 or even 15,000 reps in this book. Our laptops are fast enough and then these two numbers will be closer, and our argument will be more convincing. 

Also - we need to bring this back to what we are doing! This distributions tells us what we might expect when taking a single random sample. And if we can somehow learn how to predict what sampling distributions should look like for any sample, we can use this knowledge to report more than just a single point estimate when we report our estimation results when we are in the more common, and realistic scenario of only having a single sample.

10-inference.Rmd
```diff
@@ -0,0 +1,806 @@
...
+Timbits. What if we wanted to infer something about a population of *quantitative* variables instead?
+As mentioned in the introduction to this chapter, there are many choices of population parameter
+for each type of observed variable. In this section, we will study the case where we are interested
+in the population mean of a quantitative variable.
+
+In particular, we will look at an example using data from Airbnb, an online
+marketplace for arranging or offering places to stay. The dataset contains
+Airbnb listings for Vancouver, Canada, in September 2020 
+from [Inside Airbnb](http://insideairbnb.com/). 
+Let's imagine (for learning purposes) that our dataset represents the population of all Airbnb rental listings in Vancouver,
```
ttimbers, line 236: If this is indeed the rates for all of Vancouver AirBnBs, can we not take a stronger stance and say that this is our population?

10-inference.Rmd
```diff
@@ -0,0 +1,806 @@
...
+and we are interested in the population mean price per night.
```
ttimbers, line 237: This comes too late, let's move our question (well let's actually write an inferential question for this case) up to the third sentence of the first paragraph of this section.

10-inference.Rmd
```diff
@@ -0,0 +1,806 @@
...
+  na.omit()
+airbnb <- airbnb %>% 
+  mutate(id = 1:nrow(airbnb))
+head(airbnb)
+```
+
+We can visualize the population distribution of the price per night with a histogram. 
+```{r 11-example-means2, echo = TRUE, message = FALSE, warning = FALSE, fig.cap = "Population distribution of price per night ($) for all Airbnb listings in Vancouver, Canada"}
+population_distribution <-  ggplot(airbnb, aes(x = price)) +
+    geom_histogram(fill="#0072B2", color="#e9ecef") +
```
ttimbers, line 255: again, let's use names for colours. There is this useful PDF that can help with this: http://www.stat.columbia.edu/~tzheng/files/Rcolor.pdf

10-inference.Rmd
```diff
@@ -0,0 +1,806 @@
...
+    xlab("Price per night ($)") 
+population_distribution
```
ttimbers, line 257: We should have some narration about what we are calculating here. Also, I see no need to calculate the standard deviation...

10-inference.Rmd
```diff
@@ -0,0 +1,806 @@
...
+    summarize(pop_mean = mean(price),
+             pop_sd = sd(price))
+population_parameters
+```
+
+We see that the distribution has one peak and is skewed&mdash;most of the listings
+are less than \$250 per night, but a small proportion of listings cost more
+than that, creating a long tail on the histogram's right side. 
+The population mean is \$`r round(population_parameters$pop_mean,2)` and the
+population standard deviation is \$`r round(population_parameters$pop_sd, 2)`.
```
ttimbers, line 268: I don't think we need to calculate the sd, we don't really do anything with it...

10-inference.Rmd
```diff
@@ -0,0 +1,806 @@
...
+
+Suppose we take a sample of 20 observations from our population. Below we 
+create a histogram to visualize the
+distribution of observations in the sample,
+and calculate the mean and standard deviation of our sample. These two numbers
+are **point estimates** for the mean and standard deviation of the full population.
+
+```{r 11-example-means3, echo = TRUE, message = FALSE, warning = FALSE, fig.cap = "Distribution of price per night ($) for sample of 20 Airbnb listings"}
```
ttimbers, line 276: figure is bigger than necessary

10-inference.Rmd
```diff
@@ -0,0 +1,806 @@
...
+  select(id, neighbourhood = neighbourhood_cleansed, room_type, accommodates, bathrooms = bathrooms_text, bedrooms, beds,  price) %>% 
+  mutate(price = as.numeric(str_remove(price, "[$]"))) %>% 
+  na.omit()
+airbnb <- airbnb %>% 
+  mutate(id = 1:nrow(airbnb))
+head(airbnb)
+```
+
+We can visualize the population distribution of the price per night with a histogram. 
+```{r 11-example-means2, echo = TRUE, message = FALSE, warning = FALSE, fig.cap = "Population distribution of price per night ($) for all Airbnb listings in Vancouver, Canada"}
```
ttimbers, line 253: figure is bigger than necessary

10-inference.Rmd
```diff
@@ -0,0 +1,806 @@
...
+```{r 11-example-proportions6, echo = TRUE, message = FALSE, warning = FALSE}
+sample_estimates <- samples %>% 
+    group_by(replicate) %>% 
+    summarise(sample_proportion = sum(flavour == "chocolate") / 40)
+head(sample_estimates)
+tail(sample_estimates)
+```
+
+```{r 11-example-proportions7, echo = TRUE, message = FALSE, warning = FALSE,fig.cap = "Sampling distribution of the sample proportion for sample size 40"}
+sampling_distribution <-  ggplot(sample_estimates, aes(x = sample_proportion)) +
```
ttimbers, line 199: figure is bigger than necessary

10-inference.Rmd
```diff
@@ -0,0 +1,806 @@
...
+population_parameters
+```
+
+We see that the distribution has one peak and is skewed&mdash;most of the listings
+are less than \$250 per night, but a small proportion of listings cost more
+than that, creating a long tail on the histogram's right side. 
+The population mean is \$`r round(population_parameters$pop_mean,2)` and the
+population standard deviation is \$`r round(population_parameters$pop_sd, 2)`.
+
+Suppose we take a sample of 20 observations from our population. Below we
```
ttimbers, line 270: Why such a small sample size? It might be better/simpler to pick a larger sample size where CLT kicks in, even if we don't talk about it here.

ttimbers, line 270:
```suggestion
Suppose that we did not have access to the population data, yet we still wanted to estimate the mean price per night? We could answer this question by taking a random sample of as many AirBnB listings as we had time to, let's say we could do this for 40 listings. What would such a sample look like?

Let's take advantage of the fact that we do have access to the population data and simulate taking one random sample of 40 listings in R, again using `rep_sample_n`. After doing this we
```

10-inference.Rmd
```diff
@@ -0,0 +1,806 @@
...
+are less than \$250 per night, but a small proportion of listings cost more
+than that, creating a long tail on the histogram's right side. 
+The population mean is \$`r round(population_parameters$pop_mean,2)` and the
+population standard deviation is \$`r round(population_parameters$pop_sd, 2)`.
+
+Suppose we take a sample of 20 observations from our population. Below we 
+create a histogram to visualize the
+distribution of observations in the sample,
+and calculate the mean and standard deviation of our sample. These two numbers
+are **point estimates** for the mean and standard deviation of the full population.
```
ttimbers, line 274: again, remove sd, and just focus on one parameter to estimate?

10-inference.Rmd
```diff
@@ -0,0 +1,806 @@
...
+about `r round(100*abs(estimates$sample_sd - population_parameters$pop_sd)/population_parameters$pop_sd, 1)`% off.
+Note that in practice, we usually cannot compute the accuracy of the estimate, since we do not have access to the population
+parameter; if we did, we wouldn't need to estimate it!
+
+Also recall from the previous section that the point estimate can vary; if 
+we took another random sample from the population, then the value of our statistic may change.
+So then did we just get lucky with our point estimate above?
+How much does our estimate vary across different samples of size 20 in this example? Again, since we have access to the population,
+we can take many samples and plot the **sampling distribution** of the point estimates to get a sense
+for this variation. In this case, we'll use 1500 samples of size 20.
```
ttimbers, line 306: Again, we have fast laptops, let's up the reps a bit.

10-inference.Rmd
```diff
@@ -0,0 +1,806 @@
...
+while the standard deviation was 
+about `r round(100*abs(estimates$sample_sd - population_parameters$pop_sd)/population_parameters$pop_sd, 1)`% off.
+Note that in practice, we usually cannot compute the accuracy of the estimate, since we do not have access to the population
+parameter; if we did, we wouldn't need to estimate it!
+
+Also recall from the previous section that the point estimate can vary; if 
+we took another random sample from the population, then the value of our statistic may change.
+So then did we just get lucky with our point estimate above?
+How much does our estimate vary across different samples of size 20 in this example? Again, since we have access to the population,
+we can take many samples and plot the **sampling distribution** of the point estimates to get a sense
```
ttimbers, line 305:
```suggestion
we can take many samples and plot the **sampling distribution** of sample means for samples of size 20 to get a sense
```

10-inference.Rmd
```diff
@@ -0,0 +1,806 @@
...
+```{r 11-example-means4.5}
+sample_estimates %>% 
+  summarise(mean_of_sample_means = mean(sample_mean))
+```
+Notice that the mean of the sample means is \$`r round(mean(sample_estimates$sample_mean),2)`. Recall that the population mean
+was \$`r round(mean(airbnb$price),2)`. 
+-->
+
+```{r 11-example-means5, echo = FALSE, message = FALSE, warning = FALSE, fig.cap = "Comparision of population distribution, sample distribution and sampling distribution"}
+grid.arrange(population_distribution + ggtitle("Population distribution"),
```
ttimbers, line 344: maybe shrink the width of this panel plot? And should we add the sample size for the bottom plot, given that is only the sampling dist of the mean for that sample size?

ttimbers: Also, you need to keep the x-axes the same here. So that the plots are comparable.

10-inference.Rmd
```diff
@@ -0,0 +1,806 @@
...
+             annotated_sampling_dist_100,  
+             annotated_sampling_dist_500,  
+             nrow = 2, ncol = 2) 
+```
+
+Based on the visualization, two points about the sample mean become clear.
+First, the
+mean of the sample mean (across samples) is equal to the population mean.
+Second, increasing the size of the sample
+decreases the standard deviation (i.e., the variability) in the sample mean
```
ttimbers, line 432:
```suggestion
decreases the spread (i.e., the variability) in the sample mean
```

ttimbers:
@leem44 - I have finished the section on sampling distributions. Tomorrow I will read and comment on the bootstrap section. I think it is fine if you work on changes to address my comments in the sampling distributions section on this branch and push to GitHub here. It will just update this conversation.

I can still comment on the rest of the file here while you do this. For some of my changes I suggested, if you are OK with them, I have suggested them as commits which you can just click "Commit suggestion" to accept. I suggest you start with those, and then pull these changes into the local copy of the branch, and then work on any changes needed for my more general comments.

leem44:
@ttimbers Great thanks so much, Tiffany!!! I will look at it right now!

10-inference.Rmd
```diff
@@ -0,0 +1,789 @@
...
+population (`one_sample`)  using `rep_sample_n` and calculate the means for
+each of those replicates. Recall that this assumes that `one_sample` *looks like*
+our original population; but since we do not have access to the population itself,
+this is often the best we can do.
+
+```{r 11-bootstrapping4, echo = TRUE, message = FALSE, warning = FALSE}
+boot1500 <- one_sample %>%
+    rep_sample_n(size = 40, replace = TRUE, reps = 1500)
+head(boot1500)
+tail(boot1500)
```
ttimbers, line 576: After showing `head` and `tail`, it might be nice to show 6 or so sample distributions to show how they differ, something like this: 

Screen Shot 2020-11-19 at 12 59 34 PM

And then calculate and show the point estimates for these six, or so, samples:

Screen Shot 2020-11-19 at 12 59 48 PM

ttimbers: Source for this is here: https://pages.github.ubc.ca/MDS-2020-21/DSCI_552_stat-inf-1_students/lectures/02_lecture-sampling-and-bootstrapping.html

10-inference.Rmd
```diff
@@ -0,0 +1,789 @@
...
+
+sampling_dist <-  ggplot(sample_estimates, aes(x = sample_mean)) +
+    geom_histogram(fill="#0072B2", color="#e9ecef") +
+    xlab("Sample mean price per night ($)")
+
+sampling_dist <- sampling_dist +
+  annotate("text", x = max_x(sampling_dist), max_count(sampling_dist), hjust = 1, vjust =1,
+           label = paste("mean = ", round(mean(sample_estimates$sample_mean), 1))) + 
+  annotate("text", x = max_x(sampling_dist), max_count(sampling_dist), hjust = 1, vjust =3,
+           label = paste("sd = ", round(sd(sample_estimates$sample_mean), 1))) +
```
ttimbers, line 612: I don't think we should annotate the standard deviation. Instead we can just eyeball the spread and say that it looks similar.

10-inference.Rmd
```diff
@@ -0,0 +1,789 @@
...
+First, we see how the sampling distribution is centred at \$`r round(mean(airbnb$price),2)`, the 
+population mean value, as we learned in
+the previous section. However, the bootstrap distribution is centred at \$`r round(mean(boot1500_means$mean), 2)`, which 
+is the original sample's mean price per night. This is because we are resampling from the original sample over and
+over again, so we see that the bootstrap distribution is centred at the
+original sample's mean value. The second important point is that the shape and spread of the true sampling distribution
+and the bootstrap distribution is similar; the bootstrap distribution lets us get a sense for the variability of the 
+point estimate.
+
+```{r 11-bootstrapping7, echo = F, message = FALSE, warning = FALSE, fig.cap = "Summary of bootstrapping process"}
```
ttimbers, line 636: This diagram is really really great! I still think we should include a more toy version with numbers, like my sketch, earlier in this section when we introduce the bootstrap.

10-inference.Rmd
```diff
@@ -0,0 +1,789 @@
...
+```
+There are two important points that we can take away from these plots.
+First, we see how the sampling distribution is centred at \$`r round(mean(airbnb$price),2)`, the 
+population mean value, as we learned in
+the previous section. However, the bootstrap distribution is centred at \$`r round(mean(boot1500_means$mean), 2)`, which 
+is the original sample's mean price per night. This is because we are resampling from the original sample over and
+over again, so we see that the bootstrap distribution is centred at the
+original sample's mean value. The second important point is that the shape and spread of the true sampling distribution
+and the bootstrap distribution is similar; the bootstrap distribution lets us get a sense for the variability of the 
+point estimate.
```
ttimbers, line 634: The most important point here is that the bootstrap distribution pretty well approximates the spread of the sampling distribution. And we can use that spread to help come up with a plausible range for where we expect our population parameter's value to lie. 

A secondary note - is that unlike the sampling distribution's mean, the bootstrap distribution's mean is not the value of the population parameter - this is because all the bootstrap samples came from a single sample from the population - and so most of their means will be similar to that single sample (so the bootstrap distribution's mean, if we did enough replicates would converge on the mean of the single sample).

Here's what I used in my MDS notes this year:

What do we notice?

The shape and spread of the distribution of the bootstrap sample means approximates that of the sampling distribution of sample means!

The means of these two distributions are different: the mean of the boostrap sample means is almost exactly that of the original one sample mean, whereas the mean of the sampling distribution of sample means is almost exactly that of the population parameter we are interested in estimating.

Idea! We can use this distribution of the bootstrap sample means to approximate the sampling distribution of sample means when we only have one sample - and then we can use this to generate a plausible range where we believe the value of true population parameter lies to report along with our estimate!

10-inference.Rmd
ttimbers: The dots are a bit strange, I understand what you are trying to do here, and it makes sense, however I think it's their placement that is a bit odd...

10-inference.Rmd
```diff
@@ -0,0 +1,789 @@
...
+          y = unit(0.16, "npc"),gp = gpar(fontsize=15))
+grid.text(".", x = unit(0.7, "npc"), 
+          y = unit(0.18, "npc"),gp = gpar(fontsize=15))
+
+```
+
+### Using the bootstrap to calculate a plausible range  
+
+Now that we have constructed our bootstrap distribution let's use it to create
+an approximate bootstrap confidence interval, a range of plausible values for the population mean. We will construct a 95\% percentile bootstrap confidence interval and find the range of values that cover the middle 95\% of the bootstrap distribution. Note that there's nothing particularly special about 95\%, we could have used other confidence levels, such as 90\% or 99\%. The confidence level you choose will depend on your application, which is beyond the scope of this textbook.
```
ttimbers, line 749: We need to define a 95% ci here, or this will be jibberish to the students. 

In my MDS notes, I say this (this probably comes from Modern Dive originally, and so we need to be careful to make this our own):

What does a 95% confidence interval mean?

It means that if we were to repeat the process of sampling and calculating a 95% confidence interval many many times, 95% of the time we would expect our population parameter’s value to lie within the confidence interval.

Why 95%?

Commonly used confidence levels include 90%, 95%, and 99%. All other things being equal, higher confidence levels correspond to wider confidence intervals and lower confidence levels correspond to narrower confidence intervals. The level chosen depends on the downstream analysis and what chance you are willing to take of being wrong.

Verbally then I also add that if you have a downstream decision/application where the cost of missing the population parameter in you confidence intervals is high (e.g., impacts human health or well being) then you would tend to choose a higher confidence level.

10-inference.Rmd
leem44: Changed hex to readable names, changed binwidth to bins

10-inference.Rmd
leem44: updated scenario with motivating example

10-inference.Rmd
leem44: updated scenario with motivation

10-inference.Rmd
leem44: provided more explanation here and removed some of the statistics mentioned or provided a brief explanation

10-inference.Rmd
```diff
@@ -0,0 +1,789 @@
...
+    annotate("text", x = bounds[2], max_count(boot_est_dist), hjust = 0.5, vjust =2,
+           label = paste("97.5th percentile =", round(bounds[2], 2)))
+```
+
+Here we notice that our interval does indeed contain the true
+population mean value, \$`r round(mean(airbnb$price),2)`\! However, in
+practice, we would not know whether our interval captured the true population
+parameter or not because we usually only have a single sample, not the entire
+population.
+
```
ttimbers, line 783: Maybe add however, this is the best we can do when we only have one sample!

Let's add how to report our point estimate and confidence intervals, and revisit what a confidence intervals means. Maybe loop back to the question we asked at the beginning of this section.

Finally, perhaps end on this is only the beginning of the journey into statistical inference, and we can extend the concepts learned here to do much more than report point estimates and confidence intervals, such as hypothesis testing for differences between populations, associations between variables, etc, but what we presented here is the foundation for this... something to wrap this up nicely with a 🎁

10-inference.Rmd
```diff
@@ -0,0 +1,789 @@
...
+## Additional readings
+
+For more about statistical inference and bootstrapping, refer to pages 187-190  
+of [Introduction to Statistical Learning with Applications in R](http://www-bcf.usc.edu/~gareth/ISL/ISLR%20Seventh%20Printing.pdf) by
```
ttimbers, line 787: I think this book is not that great for this audience, Modern Dive is! And if we wanted to point them to another more traditional approach to inference, I would suggest adding OpenIntro Statistics.

ttimbers (commented):
OK @leem44 ! I am now done reviewing your chapter. Great first/second draft. With some modest tweaks to further improve it, this will be great!

10-inference.Rmd
leem44: added an explanation of head() and tail() so we can remove from other chapters

10-inference.Rmd
leem44: upped the reps to 15000, added a summary paragraph at the end of proportions section

10-inference.Rmd
leem44: moved up to first paragraph, added question we want to answer

10-inference.Rmd
leem44: added some narration, removed sd calculation

10-inference.Rmd
leem44: removed sd from this example

10-inference.Rmd
leem44: changed out.width to 70%

10-inference.Rmd
leem44: changed out.width to 70%

10-inference.Rmd
leem44: changed sample size to 40

10-inference.Rmd
leem44: removed sd

10-inference.Rmd
leem44: increased to 15000

10-inference.Rmd
leem44: changed the plot width, indicated sample size and changed x axis

10-inference.Rmd
leem44: Thanks, Tiffany! I will leave this for you 👍

10-inference.Rmd
leem44: @ttimbers Yeah, I thought about this too before, I wasn't sure if I should say that since then we would have to be super specific about defining the population since this data was scrapped in September 2020 so I guess the population is Airbnb listings available only in September 2020. I thought it would be strange since what if we want to make inference about Airbnb data in October or something so I thought it would be best to say this isn't population data. I can change it if you feel strongly, but I thought it would be more confusing if we said it was!

leem44:
@ttimbers Thank you for the feedback! 
- made all changes based on Tiffany's feedback except for updating the Timbits image and saying data is the population (as per the comments in the unresolved issues above) 
@trevorcampbell ready for you now!

Issue #49, opened 2020-06-13
Add "write_csv" as a learning goal to chapter 2


jbourak:
(As mentioned in [this issue](https://github.ubc.ca/UBC-DSCI/dsci-100-instructor/issues/108) on the `dsci-100-instructor` repo)
 - use `write_csv` to save a data frame to a `.csv` file

leem44:
done

Issue #24, opened 2019-09-09
Chapter 2 future improvements


trevorcampbell:
- more explanation of why we should care about databases (can take from worksheet 02 / tut 02)
- add URL links to the different database types in 2.6 when introduced
- more explanation of difference between dbplyr wrangling and collect followed by wrangling, i.e., you can do lots with large data in databases without having to bring huge amounts of data to your comp

Lourenzutti:
Hi @ttimbers and @trevorcampbell,

this comment is for Section 2.6.

I think it reads quite well. But there are two things we can think about this, the first part is the connecting to a specific database part. This will be different for different DBMS. Then there's the data wrangling and R functions that can be used when we have this connection. This will be the same independently for what is the underlying DBMS. 

I think this point might be clearer if you separate these two stages in your discussion. Something like "Section 2.6.1 Connecting to a database"  then you show how to connect to a SQLite using RSQLite and to Postgres using RPostgres(). Then, you show the examples you have for each database. the data wrangling and listing tables and so on. In this stages the functions will be the same whatever is the DBMS.

PR #64, opened 2020-11-29
01 reading canlang

leem44:
@ttimbers -- ready for you to review!
- removed the `state_property_vote` data and replaced it with the `can_lang` dataset 
- added figure captions 
- applied styler
- addressing comments in #24 and #49 
I didn't remove any of the `state_property_vote` datasets from the `data/` folder. Not sure if we want to store these somewhere just in case...


ttimbers:
@leem44 - I made relatively small fixes in the commits above. One thing about retrieving the data, I don't think we should redo what `canlang` already does. The files you create in `data/create_can_lang_variants.R` already exist in that package. I think we should just write the files from that package repo into the data dir using `read_csv` or install the package into the docker image and access them from the environment. Or add them to Trevor's Jupyter notebook.

ttimbers:
I'm going to do the Docker container approach in another PR. I will merge this one now as everything else has been addressed.

PR #70, opened 2020-12-14
Version control

ttimbers:
Draft of version control chapter, finally ready for feedback!


trevorcampbell:
Still working on this; but I noticed some references to JupyterLab (should be jupyter hub, no?)

Will let you know when I'm done making commits to avoid conflicts

ttimbers:
Yes, I think it should be JupyterHub to keep things simple. This chapter depends on the JupyterHub using Jupyter lab as opposed to the classic notebook interface. But I don't think we need to state that here. And I think most newer JupyterHubs being built will use Jupyter lab, not the classic notebook interface.

trevorcampbell:
OK @ttimbers I'm done editing. Two things for you to fix:

- JupyterLab vs JupyterHub
- a bunch of the images aren't showing up for me -- specifically in sections 5.5.2 through 5.8.3 I don't see any images

PR #75, opened 2020-12-15
Moving to your own machine

trevorcampbell:
@ttimbers this chapter is ready to merge *except* that there is a weird citation to some Canadian statistics thing at the end. I imagine that comes from the earlier chapters that use that data, but I don't see why a citation should appear in chapter 12...

Once you fix that I'm happy to merge.

FYI: I also significantly reorganized this chapter to avoid a lot of duplicated text.

FYI2: I also took the opportunity to standardize "Additional resources" as the final subsection title in a number of chapters


ttimbers:
OK refs are now fixed and using bookdown (via bibTeX) to do this. Currently testing install instructions on mac. Will merge once I complete this sanity check.

ttimbers:
Figured out the LaTeX to converting to PDF works to boot! Merging!

Issue #116, opened 2021-04-08
Simplify and better explain data sets


ttimbers:
Where we can, provide more information/context about the data sets (maybe in a call out box or something?). Also, make it clear where things have been simplified and why (so we can focus on the data science method we are teaching). At a minimum, we need to explicitly state that data science **cannot** be done without a deep understanding of data and domain, and that we are approaching things the way we are to teach data science, and IRL data science should not be done without a domain expert, or alternatively, it is common to practice data science in your domain of expertise.*  Go through each chapter and find where we can just have one data set. Idea: see if we can have chapter 2 only use `canlang` data sets (might not work for web scraping, but maybe there's a more related data set? Note: Chapter 4 needs multiple data sets by the way we have written it. Question: Think about clustering chapter - can we use one that we are already using? This addresses comments made by Rev C & E, but we do need to also generate a rebuttal here stating why we have chosen a rich set of data sets for this book. **MAJOR**

trevorcampbell:
One thought from me here: I really like the current way that chapters 6+7 use the breast cancer data, and chapters 8+9 use the sacramento housing data. It doesn't seem like "dataset overload" since each dataset is used over two chapters that focus on one global topic. I forget exactly where our discussions about this issue landed, but count me as a "don't change the datasets we use in ch 6-9". Of course all of the above comments re: making sure to explain and think about the data more deeply still apply.

leem44:
I agree that we shouldn't change the breast cancer or Sacramento data set, and most of the early chapters use the Canadian language data set so I won't touch those. What about the clustering and inference chapters? I remember there were some discussions around changing the clustering one to something else. @trevorcampbell @ttimbers thoughts on these chapters?

trevorcampbell:
@leem44 I think the Palmer penguins data (with the species label removed) for bill depth vs bill length would be a nice example of using clustering to discover subgroups in data. You could even add the labels back in at the very end to show that the clustering procedure found meaningful subgroups (although you'd have to be very careful to mention that if we had labels in the first place, we wouldn't need to cluster).

trevorcampbell:
[...]

PR #144, opened 2021-05-11
Replace marketing with penguins data in clustering chapter

leem44:
@ttimbers @trevorcampbell Please take a look when you can! Let me know any changes I should make! 
- Addressing #116 (will do inference changes in separate branch) specifically: 

- @leem44 I think the Palmer penguins data (with the species label removed) for bill depth vs bill length would be a nice example of using clustering to discover subgroups in data. You could even add the labels back in at the very end to show that the clustering procedure found meaningful subgroups (although you'd have to be very careful to mention that if we had labels in the first place, we wouldn't need to cluster). Maybe you can spin it as "ok based on your clustering exploration you send out some scientists and they identify 3 species, here they are!"
_Originally posted by @trevorcampbell in https://github.com/UBC-DSCI/introduction-to-datascience/issues/116#issuecomment-829429597_

I decided not to show the labels of the data at the end since I thought students might be confused, but let me know if you feel strongly otherwise

- replaced marketing example with penguins data 
- updated writing, code and images relevant to new data 
- updated references.bib


ttimbers:
If we stick with this example for the clustering chapter, I hope we can make the narrative very clear that this is different than classification. In the course, we currently make a huge emphasis that in clustering we have no labels, which is why we must cluster - which is good and important, and I think we need to keep doing this. So my worry with this data set is that we do have labels... So I wonder if that we are not extremely careful, we might send students the wrong message to the students using such a data set for our initial clustering example...

leem44:
> If we stick with this example for the clustering chapter, I hope we can make the narrative very clear that this is different than classification. In the course, we currently make a huge emphasis that in clustering we have no labels, which is why we must cluster - which is good and important, and I think we need to keep doing this. So my worry with this data set is that we do have labels... So I wonder if that we are not extremely careful, we might send students the wrong message to the students using such a data set for our initial clustering example...

Yes, I agree with you here!  I removed the labels from the data and don't make any mention of them in the narrative at all.   All I say is "The data set has been modified for use in this chapter" but I don't mention there are any species labels in the data. Let me know what you think here!

09-clustering.Rmd
```diff
@@ -148,79 +152,83 @@ In K-means clustering, we measure the quality of a cluster by its
...
 base
 ```
 
 The second step in computing the WSSD is to add up the squared distance between each point in the cluster and the cluster center.
 We use the straight-line / Euclidean distance formula that we learned about in the classification chapter.
-In the 3-observation cluster example above, we would compute the WSSD $S^2$ via
+In the `r nrow(clus1)`-observation cluster example above, we would compute the WSSD $S^2$ via
 
-$$S^2 = \left((x_1 - \mu_x)^2 + (y_1 - \mu_y)^2\right) + \left((x_2 - \mu_x)^2 + (y_2 - \mu_y)^2\right) +\left((x_3 - \mu_x)^2 + (y_3 - \mu_y)^2\right).$$
+$$S^2 = \left((x_1 - \mu_x)^2 + (y_1 - \mu_y)^2\right) + \left((x_2 - \mu_x)^2 + (y_2 - \mu_y)^2\right) +\left((x_3 - \mu_x)^2 + (y_3 - \mu_y)^2\right) \left((x_4 - \mu_x)^2 + (y_4 - \mu_y)^2\right).$$
```
trevorcampbell, line 193: This equation is too long for rendering on a single line (at least in my web browser)

09-clustering.Rmd
```diff
@@ -148,79 +152,83 @@ In K-means clustering, we measure the quality of a cluster by its
...
+We begin the K-means algorithm by picking K, and uniformly randomly assigning data to the K clusters. 
 Then K-means consists of two major steps that attempt to minimize the
 sum of WSSDs over all the clusters, i.e. the *total WSSD*:
 
 1. **Center update:** Compute the center of each cluster.
 2. **Label update:** Reassign each data point to the cluster with the nearest center.
 
 These two steps are repeated until the cluster assignments no longer change.
-For example, in the customer data example from earlier, our initialization might look like this:
+For example, in the penguin data example, our initialization might look like this:
```
trevorcampbell, line 227: i like the color scheme, but maybe it's not great for colorblind vision? pass these images through a colorblind filter to see what they look like, and if needed change to a different scheme

09-clustering.Rmd
```diff
@@ -355,42 +367,42 @@ cross-validation with some measure of model prediction error.
 Further, if K is chosen too small, then multiple clusters get grouped together;
 if K is too large, then clusters get subdivided. In both cases, we will potentially miss
 interesting structure in the data. For example, take a look below at the K-means
-clustering of our customer satisfaction and loyalty data for a number of clusters
+clustering of our penguin flipper and bill length data for a number of clusters
 ranging from 1 to 9.
```
trevorcampbell, line 380: the figure here doesn't look right. the means don't appear to be positioned correctly within their clusters

09-clustering.Rmd
```diff
@@ -514,23 +527,23 @@ unlike other functions you've seen so far (`kmeans`, `select`, etc).
 To do this we will write our map statement like this:
 
 ```
```
trevorcampbell, line 538: this fenced code block needs the "r" syntax highlighting

references.bib
```diff
@@ -80,3 +72,18 @@ @book{james2013introduction
   publisher={Springer}
 }
 
+@Manual{palmerpenguins,    
+    title = {palmerpenguins: Palmer Archipelago (Antarctica) penguin data},
```
trevorcampbell, line 76: This gets typeset as `Palmerpenguins` in the textbook, it should be `palmerpenguins`. You can achieve this by adding curly braces around anything you want to preserve the formatting of, e.g.

```
    title = {{palmerpenguins}: Palmer Archipelago (Antarctica) penguin data},
```

trevorcampbell (changes requested):

leem44:
Thanks for the comments @trevorcampbell! Merging branch!

Issue #147, opened 2021-05-12
Use direct figure references


trevorcampbell:
throughout we do a lot of "the figure above" and "the figure below", which won't work with floating figures in latex processing.

See the clustering chapter for examples of actually referencing figures.

As far as style goes -- most typesetters capitalize, so e.g. "Figure \ref[blahblah]", not "figure \ref[blahblah]"

trevorcampbell:
added to style guide

PR #149, opened 2021-05-13
Update inference data

### LINKED ISSUE #133: Ch 11
Figure 11.15 is not centered at the sample mean

leem44:
@trevorcampbell @ttimbers Ready for your feedback! 
- Replaced timbits data set with airbnb example 
- addressing #116 and  #147
- fixed #133 

If either of you has a better picture of Vancouver, please change the one I have there! I didn't include a citation to that photo since I figured one of you may have a better photo to use?

### DIFF
--- a/10-inference.Rmd
+++ b/10-inference.Rmd
@@ -56,15 +56,15 @@ In the above question, we are interested in making a conclusion about *all*
 undergraduate students in North America; this is our **population**.
 In general, the population is the complete collection of individuals or cases we are interested in studying. 
 Further, in the above question, we are interested in computing a quantity&mdash;the proportion
-of iPhone owners&mdash;based on the entire population. This is our **population parameter**.
+of iPhone owners&mdash;based on the entire population. This proportion is our **population parameter**.
 In general, a population parameter is a numerical characteristic
 of the entire population. To compute this number in the example above, we would need to ask 
-every single undergraduate in North America whether or not they own an iPhone. In practice,
+every single undergraduate in North America whether they own an iPhone. In practice,
 directly computing population parameters is often time-consuming and costly, and sometimes impossible. 
 
 A more practical approach would be to collect measurements for a **sample**: a subset of
 individuals collected from the population. We can then compute a **sample estimate**&mdash;a numerical
-characteristic of the sample&mdash;that estimates the population parameter. For example, suppose we randomly selected 100 undergraduate students across North America (the sample) and computed the proportion of those
+characteristic of the sample&mdash;that estimates the population parameter. For example, suppose we randomly selected 10 undergraduate students across North America (the sample) and computed the proportion of those
 students who own an iPhone (the sample estimate). In that case, we might suspect that that proportion is a reasonable estimate of the proportion of students who own an iPhone in the entire population. 
 
 ```{r 11-population-vs-sample, echo = FALSE, message = FALSE, warning = FALSE, fig.cap = "Population versus sample", fig.retina = 2}
@@ -85,7 +85,7 @@ The population consists of all studio apartment rentals in Vancouver, and the po
 ## Sampling distributions
 
 ### Sampling distributions for proportions
-Airbnb is an online marketplace for arranging places to stay. We will look at an example using data from [Inside Airbnb](http://insideAirbnb.com/). The data set, `airbnb`, contains  listings for Vancouver, Canada, in September 2020. Our data contains an ID number, neighbourhood, type of room, the number of people the rental accommodates, number of bathrooms, bedrooms, beds, and the price per night. 
+We will look at an example using data from [Inside Airbnb](http://insideAirbnb.com/). Airbnb is an online marketplace for arranging places to stay. The data set, `airbnb`, contains listings for Vancouver, Canada, in September 2020. Our data includes an ID number, neighbourhood, type of room, the number of people the rental accommodates, number of bathrooms, bedrooms, beds, and the price per night. 
 
 ```{r 11-example-means1, echo = FALSE, message = FALSE, warning = FALSE}
 airbnb <- read_csv("data/listings.csv") %>%
@@ -94,15 +94,18 @@ airbnb <- read_csv("data/listings.csv") %>%
   na.omit()
 airbnb <- airbnb %>%
   mutate(id = 1:nrow(airbnb))
-head(airbnb)
 ```
+
 ```{r 11-vancouver-city, echo = FALSE, message = FALSE, warning = FALSE, fig.cap = "Vancouver, BC Canada", fig.retina = 2, out.width="880"}
 knitr::include_graphics("https://upload.wikimedia.org/wikipedia/commons/7/76/Vancouver_City_2_by_ajithrajeswari.jpg")
 ```
 
-Suppose the city of Vancouver wants to know how many entire homes and apartments are being rented out as short-term rentals on Airbnb (rather than just as private or shared rooms). Therefore they may want to estimate the true proportion of all Airbnb listings where the type of place is listed as entire home or apartment. Now, of course, usually we do not have access to the true population, but here let's imagine (for learning purposes) that our data set represents the population of all Airbnb rental listings in Vancouver, Canada. We can find the proportion of listings where `room_type == "Entire home/apt"`.
+```{r}
+airbnb
+```
+Suppose the city of Vancouver wants to know how many Airbnb places are rented out as entire homes and apartments rather than as private or shared rooms. Therefore they may want to estimate the true proportion of all Airbnb listings where the "type of place" is listed as "entire home or apartment." Of course, we usually do not have access to the true population, but here let's imagine (for learning purposes) that our data set represents the population of all Airbnb rental listings in Vancouver, Canada. We can find the proportion of listings where `room_type == "Entire home/apt"`.
 
-```{r 11-example-proportions2, echo = TRUE, message = FALSE, warning = FALSE}
+```{r 11-example-proportions2, echo = TRUE, message = FALSE, warning = FALSE }
 library(tidyverse)
 library(infer)
 
@@ -121,21 +124,22 @@ population_proportion <- airbnb %>%
   ) %>% 
   pull()
 ```
-From our data, we can see that the proportion of `Entire home/apt` listings in the data set is `r round(population_proportion,2)`. This value, `r round(population_proportion,2)`, is the *population parameter*. Remember this parameter value is usually unknown in real data analysis problems.
 
-What would happen if we were to randomly-select 40 listings and count the number of entire home/apartment listings (*i.e.,* take a random sample of size 40 from our population)? Let's use R to simulate this using our `airbnb` population. We can do this using the `rep_sample_n` function from the `infer` package. The arguments of `rep_sample_n` are (1) the data frame (or tibble) to sample from, and (2) the size of the sample to take.
+From our data, we can see that the proportion of `Entire home/apt` listings in the data set is `r round(population_proportion,2)`. This value, `r round(population_proportion,2)`, is the *population parameter*. Remember, this parameter value is usually unknown in real data analysis problems.
+
+What would happen if we were to randomly select 40 listings and count the number of entire home/apartment listings (*i.e.,* take a random sample of size 40 from our population)? Let's use R to simulate this using our `airbnb` population. We can do this using the `rep_sample_n` function from the `infer` package. The arguments of `rep_sample_n` are (1) the data frame (or tibble) to sample from, and (2) the size of the sample to take.
 
 ```{r 11-example-proportions3, echo = TRUE, message = FALSE, warning = FALSE}
-set.seed(1)
-samples_1 <- rep_sample_n(tbl = airbnb, size = 40)
-airbnb_sample_1 <- summarize(samples_1,
+set.seed(123)
+sample_1 <- rep_sample_n(tbl = airbnb, size = 40)
+airbnb_sample_1 <- summarize(sample_1,
   n = sum(room_type == "Entire home/apt"),
   prop = sum(room_type == "Entire home/apt") / 40
 )
 airbnb_sample_1
 ```
 Here we see that the proportion of entire home/apartment listings in this random sample is
-`r round(airbnb_sample_1$prop,2)`. This value is our estimate &mdash; our best guess of our population parameter using this sample. Given that it is a single
+`r round(airbnb_sample_1$prop,2)`. This value is our *estimate* &mdash; our best guess of our population parameter using this sample. Given that it is a single
 value that we are estimating, we often refer to it as a **point estimate**.
 
 Now imagine we took another random sample of 40 listings from the population. Do you
@@ -143,9 +147,9 @@ think we would get the same proportion? Let's try sampling from the population
 again and see what happens. 
 
 ```{r 11-example-proportions4, echo = TRUE, message = FALSE, warning = FALSE}
-set.seed(123)
-samples_2 <- rep_sample_n(airbnb, size = 40)
-airbnb_sample_2 <- summarize(samples_2,
+set.seed(1234)
+sample_2 <- rep_sample_n(airbnb, size = 40)
+airbnb_sample_2 <- summarize(sample_2,
   n = sum(room_type == "Entire home/apt"),
   prop = sum(room_type == "Entire home/apt") / 40
 )
@@ -159,8 +163,8 @@ different result. Estimates vary from sample to sample
 due to **sampling variability**. 
 
 But just how much should we expect the estimates of our random
-samples to vary? In order to understand this, we will simulate taking more samples
-of size 40 from our population of listings, and calculate the 
+samples to vary? To understand this, we will simulate taking more samples
+of size 40 from our population of listings and calculate the 
 proportion of entire home/apartment listings in each sample. We can then
 visualize the distribution of sample proportions we calculate. The distribution
 of the estimate for all possible samples of a given size (which we commonly refer to as $n$) from a population is
@@ -176,7 +180,7 @@ samples
 tail(samples)
 ```
 
-Notice the column `replicate` is indicating the replicate, or sample, with which each
+Notice the column `replicate` indicates the replicate, or sample, with which each
 listing belongs. Since we took 20,000 samples of size 40, there are 20,000 replicates.
 Now that we have taken 20,000 samples, to create a sampling distribution of sample proportions for samples of size 40, we need to calculate the proportion of entire home/apartment listings for each sample, $\hat{p}_\text{entire home/apartment}$:
 ```{r 11-example-proportions6, echo = TRUE, message = FALSE, warning = FALSE}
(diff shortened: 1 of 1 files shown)


trevorcampbell:
@leem44 just FYI: I rebased this branch on `dev` since this branch had that weird `x2` error in it from old `dev`

10-inference.Rmd
```diff
@@ -313,7 +303,7 @@ when we estimated the population mean
 with only `r round(100*abs(estimates$sample_mean - population_parameters$pop_mean)/population_parameters$pop_mean, 1)`% error.
 Let's visualize the population
```
trevorcampbell, line 304: why? this figure is just sort of put here without any explanation

i should say that i like the idea of including something like this, but it needs to be very carefully explained. probably an entire paragraph of text dedicated to it

10-inference.Rmd
```diff
@@ -54,15 +58,15 @@ In the above question, we are interested in making a conclusion about *all*
...
+every single undergraduate in North America whether they own an iPhone. In practice,
 directly computing population parameters is often time-consuming and costly, and sometimes impossible. 
 
 A more practical approach would be to collect measurements for a **sample**: a subset of
 individuals collected from the population. We can then compute a **sample estimate**&mdash;a numerical
-characteristic of the sample&mdash;that estimates the population parameter. For example, suppose we randomly selected 100 undergraduate students across North America (the sample) and computed the proportion of those
+characteristic of the sample&mdash;that estimates the population parameter. For example, suppose we randomly selected ten undergraduate students across North America (the sample) and computed the proportion of those
 students who own an iPhone (the sample estimate). In that case, we might suspect that that proportion is a reasonable estimate of the proportion of students who own an iPhone in the entire population. 
 
 ```{r 11-population-vs-sample, echo = FALSE, message = FALSE, warning = FALSE, fig.cap = "Population versus sample", fig.retina = 2}
```
trevorcampbell, line 72: github won't let me leave a comment further down, so i'll put it here

(around line 82 ish)

"And the list of population params we might want to calculate goes on"  -- awk sentence. could probably just remove.

10-inference.Rmd
```diff
@@ -83,124 +87,120 @@ The population consists of all studio apartment rentals in Vancouver, and the po
...
-virtual_box <- tibble(
-  timbit_id = seq(1, 10000, by = 1),
-  flavour = factor(rbinom(10000, 1, 0.63),
-    labels = c("old fashioned", "chocolate")
-  )
-)
+```{r}
+airbnb
 ```
+Suppose the city of Vancouver wants to know how many Airbnb places are listed as entire homes and apartments (rather than as private or shared rooms). Therefore they may want to estimate the true proportion of all Airbnb listings where the "type of place" is listed as "entire home or apartment." Of course, we usually do not have access to the true population, but here let's imagine (for learning purposes) that our data set represents the population of all Airbnb rental listings in Vancouver, Canada. We can find the proportion of listings where `room_type == "Entire home/apt"`.
```
trevorcampbell, line 108: why would vancouver want to know this? would be good to have some motivation

10-inference.Rmd
```diff
@@ -83,124 +87,120 @@ The population consists of all studio apartment rentals in Vancouver, and the po
...
+  summarize(
+    n =  sum(room_type == "Entire home/apt"),
+    proportion = sum(room_type == "Entire home/apt") / nrow(airbnb)
+  ) %>% 
+  pull()
+```
+
+From our data, we can see that the proportion of `Entire home/apt` listings in the data set is `r round(population_proportion,3)`. This value, `r round(population_proportion,3)`, is the *population parameter*. Remember, this parameter value is usually unknown in real data analysis problems.
+
+What would happen if we were to randomly select 40 listings and count the number of entire home/apartment listings (*i.e.,* take a random sample of size 40 from our population)? Let's use R to simulate this using our `airbnb` population. We can do this using the `rep_sample_n` function from the `infer` package. The arguments of `rep_sample_n` are (1) the data frame (or tibble) to sample from, and (2) the size of the sample to take.
```
trevorcampbell, line 132: A lot of this chapter (not just here) is kind of unmotivated like this:

"What would happen if I do X?"

there really isn't a storyline here, just a sequence of what ifs. 

I would suggest a change like the following:

 move the "Remember this param value isn't known in data analysis" from the paragraph before to start this paragraph. like "yeah that's neat but we can't compute it. so instead we are forced to do something else. maybe we can approximate with a small number of data? let's try that."

10-inference.Rmd
```diff
@@ -83,124 +87,120 @@ The population consists of all studio apartment rentals in Vancouver, and the po
...
+airbnb_sample_1
 ```
-Here we see that the proportion of chocolate Timbits in this random sample is
-`r round(choc_sample_1$prop,2)`. This value is our estimate &mdash; our best guess of our population parameter using this sample. Given that it is a single
+Here we see that the proportion of entire home/apartment listings in this random sample is
+`r round(airbnb_sample_1$prop,2)`. This value is our *estimate* &mdash; our best guess of our population parameter using this sample. Given that it is a single
 value that we are estimating, we often refer to it as a **point estimate**.
 
-Now imagine we took another random sample of 40 Timbits from the population. Do you
+Now imagine we took another random sample of 40 listings from the population. Do you
```
trevorcampbell, line 147: Again, just another "what if"

instead, motivate like so:

"Our point estimate is 0.7. wow, that's close to our true population value. but remember, we just used a random sample of 40 to estimate it. keyword there is random -- if we were to take another random sample, we wouldn't get the same answer:"

[code and result]

"confirmed! we don't get the same answer. bummer. so that means that our point estimate might be unreliable! that is sampling variability etc etc"

10-inference.Rmd
```diff
@@ -83,124 +87,120 @@ The population consists of all studio apartment rentals in Vancouver, and the po
...
 ```
 
 Notice that we get a different value for our estimate this time. The
-proportion of chocolate Timbits in this sample is `r round(choc_sample_2$prop, 2)`. 
+proportion of entire home/apartment listings in this sample is `r round(airbnb_sample_2$prop, 3)`. 
 If we were to do this again, another random sample could also give a
 different result. Estimates vary from sample to sample 
 due to **sampling variability**. 
 
 But just how much should we expect the estimates of our random
```
trevorcampbell, line 167: this is a great motivation for this bit!

10-inference.Rmd
```diff
@@ -83,124 +87,120 @@ The population consists of all studio apartment rentals in Vancouver, and the po
...
-samples to vary? In order to understand this, we will simulate taking more samples
-of size 40 from our population of Timbits, and calculate the 
-proportion of chocolate Timbits in each sample. We can then
+samples to vary? To understand this, we will simulate taking more samples
+of size 40 from our population of listings and calculate the 
+proportion of entire home/apartment listings in each sample. We can then
 visualize the distribution of sample proportions we calculate. The distribution
 of the estimate for all possible samples of a given size (which we commonly refer to as $n$) from a population is
 called a **sampling distribution**. The sampling distribution will help us see
```
trevorcampbell, line 173: should emphasize here that in the real world, since we dont have our pop, we can't actually construct the sampling distribution. but since in this eg for learning we do have it, we can actually visualize it directly.

10-inference.Rmd
```diff
@@ -210,41 +210,29 @@ the mean of the sample proportions.
...
 
-So what can we learn from this sampling distribution? This distribution tells us what we might expect from proportions from samples of size $40$ when our population proportion is 0.63. In practice, we usually don't know the proportion of our population, but if we can use what we know about the sampling distribution, we can use it to make inferences about our population when we only have a single sample. 
+So what can we learn from this sampling distribution? This distribution tells us what we might expect from proportions from samples of size $40$ when our population proportion is `r round(population_proportion,3)`. In practice, we usually don't know the proportion of our population, but if we can use what we know about the sampling distribution, we can use it to make inferences about our population when we only have a single sample.
```
trevorcampbell, line 215: "So what can we learn from this sampling distribution?"

isn't the text right before this telling us what we can learn from this sampling distribution?

"this distribution tells us what we might expect from proportions from samples of size 40 when ...."

that sentence doesn't really convey anything about what we might learn from the sampling distribution.

Honestly I might just suggest removing this paragraph, it doesn't do much

10-inference.Rmd
```diff
@@ -210,41 +210,29 @@ the mean of the sample proportions.
 sample_estimates %>%
   summarise(mean = mean(sample_proportion))
 ```
-We notice that the sample proportions are centred around the population proportion value, `r round(mean(sample_estimates$sample_proportion), 2)`! In general, the mean of the distribution of $\hat{p}$ should be equal to $p$, which is good because that means the sample proportion is neither an overestimate nor an underestimate of the population proportion.
+We notice that the sample proportions are centred around the population proportion value, `r round(mean(sample_estimates$sample_proportion), 3)`! In general, the mean of the distribution of $\hat{p}$ should be equal to $p$, which is good because that means the sample proportion is neither an overestimate nor an underestimate of the population proportion.
```
trevorcampbell, line 213: I would expand a bit to clarify the statement that "sample proportion is neither an overestimate or underestimate"

10-inference.Rmd
```diff
@@ -210,41 +210,29 @@ the mean of the sample proportions.
...
-airbnb <- read_csv("data/listings.csv") %>%
-  select(id, neighbourhood = neighbourhood_cleansed, room_type, accommodates, bathrooms = bathrooms_text, bedrooms, beds, price) %>%
-  mutate(price = as.numeric(str_remove(price, "[$]"))) %>%
-  na.omit()
-airbnb <- airbnb %>%
-  mutate(id = 1:nrow(airbnb))
-head(airbnb)
-```
+In the previous section, our variable of interest&mdash; `room_type` &mdash;was
+*categorical*, and the population parameter of interest was the proportion of listings where the room type was "entire home/apt". As mentioned in the chapter introduction, there are many choices of the population parameter for each type of observed variable. What if we wanted to infer something about a population of *quantitative* variables instead? For instance, a traveller visiting Vancouver, BC, may wish to know about the prices of staying somewhere using Airbnb. Particularly, they might be interested in estimating the population mean price per night of Airbnb listings in Vancouver, BC. This section will study the case where we are interested in the population mean of a quantitative variable.
```
trevorcampbell, line 223: again, motivation is missing. traveller might want to know avg prices why? e.g. maybe they can tell whether a particular listing is overpriced or something like that. or maybe city of van needs to know it for some reason (setting tax or something? not sure).

10-inference.Rmd
```diff
@@ -83,124 +87,120 @@ The population consists of all studio apartment rentals in Vancouver, and the po
...
+```{r 11-population-parameter, echo = F}
+population_proportion <- airbnb %>%
+  summarize(
+    n =  sum(room_type == "Entire home/apt"),
+    proportion = sum(room_type == "Entire home/apt") / nrow(airbnb)
+  ) %>% 
+  pull()
+```
+
+From our data, we can see that the proportion of `Entire home/apt` listings in the data set is `r round(population_proportion,3)`. This value, `r round(population_proportion,3)`, is the *population parameter*. Remember, this parameter value is usually unknown in real data analysis problems.
```
trevorcampbell, line 130: inconsistency here whether we emphasize important terms with italics or boldface. i think the rest of the chapter uses bold

10-inference.Rmd
```diff
@@ -256,42 +244,43 @@ population_parameters
 ```
 The price per night of all Airbnb rentals in Vancouver, BC is \$`r round(population_parameters$pop_mean,2)`, on average. This value is our population parameter since we are calculating it using the population data.
```
trevorcampbell, line 245: probably should bold important terms (population parameter) again to mimic the previous section. seems like we do that below with point estimate already

10-inference.Rmd
```diff
@@ -438,7 +426,7 @@ grid.arrange(annotated_sampling_dist_20,
 )
 ```
 
-Based on the visualization, two points about the sample mean become clear.
```
trevorcampbell, line 441: I think we need to use the term "sampling distribution" in this paragraph to emphasize to the students that what they're seeing is different sampling distributions, one for each size of sample

10-inference.Rmd
```diff
@@ -454,7 +442,7 @@ in a more reliable point estimate of the population parameter.
 ### Summary
 1. A *point estimate* is a single value computed using a sample from a population (e.g. a mean or proportion)
 2. The *sampling distribution* of an estimate is the distribution of the estimate for all possible samples of a fixed size from the same population.
-3. The sample means and proportions calculated from samples are centered around the population mean and proportion, respectively.
+3. The sample means and proportions calculated from samples are centred around the population mean and proportion, respectively.
 4. The spread of the sampling distribution is related to the sample size. As the sample size increases, the spread of the sampling distribution decreases. 
 5. The shape of the sampling distribution is usually bell-shaped with one peak and centred at the population mean or proportion.
```
trevorcampbell, line 448: The paragraph below (Why all this emphasis on...?) needs heavy editing to flow better. Lots of internal redundancy, and redundancy with the text at the start of bootstrapping.

Since it's basically a motivation for bootstrapping, I would strongly suggest that we remove it from this section and merge it with the text at the beginning of 11.5.

10-inference.Rmd
```diff
@@ -601,7 +577,7 @@ boot1_dist
 summarise(boot1, mean = mean(price))
 ```
 
-Notice that our bootstrap distribution has a similar shape to the original
+Notice in Figure \@ref(fig:11-bootstrapping3) that our bootstrap distribution has a similar shape to the original
```
trevorcampbell, line 580: again 15K here -- not a big deal but we should make them all have the same number of replicates for consistency

10-inference.Rmd
```diff
@@ -650,6 +626,7 @@ tail(boot15000_means)
 boot_est_dist <- ggplot(boot15000_means, aes(x = mean)) +
   geom_histogram(fill = "dodgerblue3", color = "lightgrey") +
   xlab("Sample mean price per night ($)")
+boot_est_dist
 ```
 
 Let's compare our bootstrap distribution with the true sampling distribution (taking many samples from the population).
```
trevorcampbell, line 632: When putting two plots together for comparison, their axes should be the same. Here the sampling distribution plot looks like it goes from 90 to 300, while the bootstrap dist goes from 100 to 210. Hard to compare two distributions if their axes are different.

Same comment for all groups of plots throughout this chapter.

10-inference.Rmd
```diff
@@ -822,22 +802,11 @@ grid.text("many means...",
...
...for the population mean. We will build a 95\% percentile bootstrap confidence interval and find the range of values that cover the middle 95\% of the bootstrap distribution. A 95\% confidence interval means that if we were to repeat the sampling process and calculate 95\% confidence intervals each time and repeat this process many times, then 95\% of the intervals would capture the population parameter's value. Note that there's nothing particularly special about 95\%, we could have used other confidence levels, such as 90\% or 99\%. There is a balance between our level of confidence and precision. A higher confidence level corresponds to a wider range of the interval, and a lower confidence level corresponds to a narrower range. Therefore the level we choose is based on what chance we are willing to take of being wrong based on the implications of being wrong for our application. In general, we choose confidence levels to be comfortable with our level of uncertainty, but not so strict that the interval is unhelpful. For instance, if our decision impacts human life and the implications of being wrong are deadly, we may want to be very confident and choose a higher confidence level.
```
trevorcampbell, line 809: uses "confidence interval" when defining of "confidence interval"

> A 95% confidence interval means that if we were to repeat the sampling process and calculate 95% confidence intervals each time and repeat this process many times, then 95% of the intervals would capture the population parameter’s value

this sentence is super long and hard to parse.

I would say in general this paragraph is tough to read -- it needs a round of careful editing

10-inference.Rmd
```diff
@@ -870,7 +839,7 @@ boot_est_dist +
   )
 ```
 
-To finish our estimation of the population parameter, we would report the point estimate and our confidence interval's lower and upper bounds. Here the sample mean price-per-night of 40 Airbnb listings was \$`r round(mean(sample_1$price),2)`, and we are 95\% "confident" that the true population mean price-per-night for all Airbnb listings in Vancouver is between \$(`r round(bounds[1],2)`, `r round(bounds[2],2)`).
+To finish our estimation of the population parameter, we would report the point estimate and our confidence interval's lower and upper bounds. Here the sample mean price-per-night of 40 Airbnb listings was \$`r round(mean(one_sample$price),2)`, and we are 95\% "confident" that the true population mean price-per-night for all Airbnb listings in Vancouver is between \$(`r round(bounds[1],2)`, `r round(bounds[2],2)`).
 
 Notice that our interval does indeed contain the true
 population mean value, \$`r round(mean(airbnb$price),2)`\! However, in
```
trevorcampbell, line 845: below in the paragraph "this chapter is only the beginning"

undefined technical terms here (e.g. "hypothesis testing") -- I would suggest just describing in plain english what we can do beyond this chapter

trevorcampbell (commented):
@leem44 The switch from timbits to airbnb looks good to me (modulo some comments about motivation that you'll see below)!

Since I was going through in detail, I did leave a bunch of comments to address regarding writing etc. May as well address them now since there aren't any other PRs for inference open.

10-inference.Rmd
```diff
@@ -83,124 +87,120 @@ The population consists of all studio apartment rentals in Vancouver, and the po
...
+airbnb <- read_csv("data/listings.csv") %>%
+  select(id, neighbourhood = neighbourhood_cleansed, room_type, accommodates, bathrooms = bathrooms_text, bedrooms, beds, price) %>%
+  mutate(price = as.numeric(str_remove(price, "[$]"))) %>%
+  na.omit()
+airbnb <- airbnb %>%
+  mutate(id = 1:nrow(airbnb))
 ```
 
+```{r 11-vancouver-city, echo = FALSE, message = FALSE, warning = FALSE, fig.cap = "Vancouver, BC Canada", fig.retina = 2, out.width="880"}
+knitr::include_graphics("https://upload.wikimedia.org/wikipedia/commons/7/76/Vancouver_City_2_by_ajithrajeswari.jpg")
```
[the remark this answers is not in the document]
leem44: Thats true! not necessary -- removing!

10-inference.Rmd
```diff
@@ -83,124 +87,120 @@ The population consists of all studio apartment rentals in Vancouver, and the po
...
 samples
 tail(samples)
 ```
 
-Notice the column `replicate` is indicating the replicate, or sample, with which each
-Timbit belongs. Since we took 15,000 samples of size 40, there are 15,000 replicates.
-Now that we have taken 15,000 samples, to create a sampling distribution of sample proportions for samples of size 40, we need to calculate the proportion of chocolate Timbits for each sample, $\hat{p}_\text{chocolate}$:
+Notice the column `replicate` indicates the replicate, or sample, with which each
+listing belongs. Since we took 20,000 samples of size 40, there are 20,000 replicates.
+Now that we have taken 20,000 samples, to create a sampling distribution of sample proportions for samples of size 40, we need to calculate the proportion of entire home/apartment listings for each sample, $\hat{p}_\text{entire home/apartment}$:
```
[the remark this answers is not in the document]
leem44: good point! I don't think its entirely necessary to define. I removed it!

10-inference.Rmd
leem44: good point! I added a short explanation

10-inference.Rmd
leem44: I added something about using the information to city bylaws -- I don't know if that is realistic or not... Feel free to let me know if I should change that!

10-inference.Rmd
leem44: Good point! I added the line: "Any *individual* sample may give you a 
sample proportion that is larger or smaller than the population proportion, but when we repeat the sampling process many times (as we did in this example), there is no tendency towards over or underestimating." Hopefully that is better!

10-inference.Rmd
leem44: Removed and merged with the section below! And checked for redundancy

Issue #142, opened 2021-05-07
Chapter 4: add nice scatter plot and need for axis scaling example


ttimbers:
Move the scatter plot example pictured below from chapter 1 as an example in chapter 4 of scatter plots and need for axis scaling. We will need to tighten up the current narrative and but aim to keep the iterative "build up" of the figure.

<img width="905" alt="Screen Shot 2021-05-07 at 12 33 42 PM" src="https://user-images.githubusercontent.com/6865016/117499733-7ae43800-af30-11eb-8b91-88b5a92c2dc6.png">

trevorcampbell:
a few adjustments too:

- the final viz should separate english and french with different colours
- we need to check over all the claims that we make to make sure they're true. For example, right after Fig 1.6, we say "From the visualization above, we can now clearly see that not just a lot, but that the majority of Canadians **reported English as their mother tongue**" but this isn't clear from the figure since none of the languages are coloured yet

Issue #130, opened 2021-04-08
Add vector data types explanations


ttimbers:
Add base R vector types (logical, integer, double, character) introduction and explanation to section 3.3.2 ("What is a vector?") in chapter 3. Add factor vector type explanation in chapter 4 (visualization chapter), when we need it. This will address comments made by Rev C & E. **MINOR**

leem44:
addressed here: #143

PR #150, opened 2021-05-14
Ch4 scatterplot

leem44:
@ttimbers @trevorcampbell Let me know any changes I should make! 
Addressing #142, #147,  #130
- moved `canlang` scatterplot example from chapter 1 to viz chapter 
-  added explanation of factors 
- updated any reference to figures 

I was thinking we could remove the old faithful example since we now have another example of scatterplots, but let me know what you think and if you agree I can remove it!


03-viz.Rmd
```diff
@@ -290,6 +290,151 @@ faithful_scatter <- ggplot(faithful, aes(x = waiting, y = eruptions)) +
 faithful_scatter
 ```
 
+
+### The Canadian languages data set
+
+We will return to the `can_lang` data set, which contains counts of languages from the 2016 Canadian census.
```
ttimbers, line 296: We should cite the census and `canlang` again here in case anyone jumps into reading the textbook here. We should also formally cite all the data in this chapter.

03-viz.Rmd
```diff
@@ -214,7 +214,7 @@ islands_bar <- ggplot(islands_df, aes(x = landmass, y = size)) +
 islands_bar
 ```
 
-Alright, not bad! This plot is definitely the right kind of visualization, as we can clearly see and compare sizes of landmasses. The major issues are that the smaller landmasses' sizes are hard to distinguish, and the names of the landmasses are obscuring each other as they have been squished into too little space. But remember that the question we asked was only about the largest landmasses; let's make the plot a little bit clearer by keeping only the largest 12 landmasses. We do this using the `top_n` function.
+Alright, not bad! The plot in Figure \@ref(fig:03-data-islands-bar) is definitely the right kind of visualization, as we can clearly see and compare sizes of landmasses. The major issues are that the smaller landmasses' sizes are hard to distinguish, and the names of the landmasses are obscuring each other as they have been squished into too little space. But remember that the question we asked was only about the largest landmasses; let's make the plot a little bit clearer by keeping only the largest 12 landmasses. We do this using the `top_n` function.
```
ttimbers, line 217: From what I have read about data visualization (see: https://clauswilke.com/dataviz/proportional-ink.html#visualizations-along-linear-axes), we often do not want to start our axes where `ggplot`'s default puts them, as it inflates differences. We should address this here and show them how to change the axes to force them to start from 0.

03-viz.Rmd
```diff
@@ -290,6 +290,151 @@ faithful_scatter <- ggplot(faithful, aes(x = waiting, y = eruptions)) +
...
+ggplot(can_lang, aes(x = most_at_home, y = mother_tongue)) +
+  geom_point()
+``` 
+To improve the interpretability of the data visualization in Figure \@ref(fig:mother-tongue-vs-most-at-home) we should replace the default axis names with a more informative labels. 
+
+```{r mother-tongue-vs-most-at-home-labs, fig.width=5.75, fig.height=5, warning=FALSE, fig.cap = "Scatter plot of number of Canadians reporting a language as their mother tongue vs the primary language at home with x and y labels"}
+ggplot(can_lang, aes(x = most_at_home, y = mother_tongue)) +
+  geom_point() +
+  xlab("Language spoken most at home (number of Canadian residents)") +
+  ylab("Mother tongue (number of Canadian residents)")
```
ttimbers, line 317: Can we teach them about using `/n` in labels to create line breaks? It is useful and makes the axes labels on these plots more readable.

03-viz.Rmd
```diff
@@ -290,6 +290,151 @@ faithful_scatter <- ggplot(faithful, aes(x = waiting, y = eruptions)) +
...
+```
+
+Most of the data points from the 214 observations in this data set are bunched up in the lower left-handside of the visualization in Figure \@ref(fig:mother-tongue-vs-most-at-home-labs). The data is clumped because many more people in Canada speak the two languages represented by the points in the upper right corner. By filtering the data, we can confirm these two points correspond to the two official languages (English and French):
+
+```{r english-mother-tongue}
+can_lang %>%
+  filter(language == "English" | language == "French")
+```
+> `|` is a logical operator representing "OR", which allows us to filter the rows where the language column is equal to "English" OR equal to "French".
```
ttimbers, line 326: Do we not teach them about how filter uses and and or in wrangling? If not, we should add this. Essentially, they need to know that `filter` treats `,` in it's call as and's, and the rows must satisfy all conditions. And then we should teach them to use `|` if they want an or. Then we can remove this note from here.

03-viz.Rmd
```diff
@@ -290,6 +290,151 @@ faithful_scatter <- ggplot(faithful, aes(x = waiting, y = eruptions)) +
...
+
+Most of the data points from the 214 observations in this data set are bunched up in the lower left-handside of the visualization in Figure \@ref(fig:mother-tongue-vs-most-at-home-labs). The data is clumped because many more people in Canada speak the two languages represented by the points in the upper right corner. By filtering the data, we can confirm these two points correspond to the two official languages (English and French):
+
+```{r english-mother-tongue}
+can_lang %>%
+  filter(language == "English" | language == "French")
+```
+> `|` is a logical operator representing "OR", which allows us to filter the rows where the language column is equal to "English" OR equal to "French". 
+
+To answer our question, we will need to adjust the scale of the x and y axes so that they are on a log scale:
```
ttimbers, line 328: We should explain here how a log scale will solve our problem of the bunched up data points, as well as the gotcha's associated with using it (in particular, values of 0...)

03-viz.Rmd
```diff
@@ -290,6 +290,151 @@ faithful_scatter <- ggplot(faithful, aes(x = waiting, y = eruptions)) +
...
+
+```{r changing_the_units, include = FALSE}
+english_mother_tongue <- can_lang %>%
+  filter(language == "English") %>%
+  pull(mother_tongue)
+census_popn <- 35151728
+```
+
+We saw that `r format(english_mother_tongue, scientific = FALSE, big.mark = ",")` people reported that their mother tongue was English in the 2016 Canadian census, but what does this number really mean? To really understand this number, we need context. In particular, how many people were in Canada when this data was collected? From the 2016 Canadian census profile, we can see that the population was reported to be `r format(census_popn, scientific = FALSE, big.mark = ",")` people. The number of people who report that English is their mother tongue is much more meaningful when we report it in this context. We can even go a step further and transform this count to a relative frequency or proportion to represent this as a single meaningful number in our data visualizations. We can do this by dividing the number of people reporting a given language as their mother tongue by the number of people who live in Canada. For example, the proportion of people who reported that their mother tongue was English in the
```
ttimbers, line 348: Can we do percentage instead of proportion? I think that is generally more understandable to a wider audience, especially when the proportion values are so small as they are here.

ttimbers (commented):
For the `canlang` scatter plot, I think we need to up the narrative on the data visualization aspects. We could reduce the narrative on the data wrangling since we have now taught that in the previous chapter. Specifically, we learn the following new things about scatter plots from this example:

1. transforming axes

2. mapping a categorical column to a colour and shape. We could move the discussion of colour-blind friendly palettes from the worksheets/later chapters to here. 

These new things, make this a more complex scatter plot and I think its great to have the simple Old Faithful one first, followed by this one.

--

Also, a couple big picture things I think we need to be changed for this chapter (which maybe should be added to an issue specifically for this chapter, instead of this global issue PR) include:

1. I think the organization of plotting stories should be grouped under plot types, we currently bounce back and forth between scatter and bar charts... Let's do line charts, then scatter, then bar charts? (or we could swap bar charts for scatter, either would be fine with me).

2. The grey background on the plots is not great. Can we teach them the use `theme` to remove it and have a plot with a white background?

03-viz.Rmd
```diff
@@ -346,7 +489,10 @@ morley_hist <- ggplot(morley, aes(x = Speed, fill = factor(Expt))) +
 morley_hist
 ```
 
-Unfortunately, the attempt to separate out the experiment number visually has created a bit of a mess. All of the colours
+
+> Notice that we wrote `factor(Expt)` in the' fill' aesthetic mapping. A **factor** helps us represent categorical data. Factors can be ordered or unordered, and even though factors look like characters, they are integers underneath. They contain a set of values, known as **levels**. Storing something as a factor helps ensure that R will treat the data correctly. For instance, if we look at the printed `morley` data, the column `Expt` is an integer (we see the label `<int>` underneath the `Expt` column name). But, we want to treat it as a categorical variable, so we use the function `factor()` to make sure it is treated as such.
```
ttimbers, line 493: I think factors should be explained/introduced in the wrangling chapter with the other vector types, and then rediscussed here. In particular, they impact plots two ways - ensuring a colour (or other channel) is mapped discretely instead of qualitatively when appropriate, as well as the ordering of things in a plot. `reorder` and `arrange` have quite different influences on a plot (with the former having a lot, and the latter having almost none).

leem44:
Moved the comments below to #170:
> --
> 
> Also, a couple big picture things I think we need to be changed for this chapter (which maybe should be added to an issue specifically for this chapter, instead of this global issue PR) include:
> 
> 1. I think the organization of plotting stories should be grouped under plot types, we currently bounce back and forth between scatter and bar charts... Let's do line charts, then scatter, then bar charts? (or we could swap bar charts for scatter, either would be fine with me).
> 2. The grey background on the plots is not great. Can we teach them the use `theme` to remove it and have a plot with a white background?
> 3. Should we teach densities here? We use them in one worksheet/tutorial in the course but there is currently nothing about them here...

03-viz.Rmd
leem44: @ttimbers It looks like this barplot does start at 0? Could you clarify this point? thanks!

leem44:
@ttimbers Thank you for your feedback!!! I made a new issue for the other "non-global" issue that you mentioned above. Other than that I resolved everything else!

Issue #84, opened 2020-12-16
LaTeX packages to add to Docker image


ttimbers:
To build the book in PDF, we need to add these to the Docker image:

```
amsmath
latex-amsmath-dev
fontspec
tipa
unicode-math
xunicode
kvoptions
ltxcmds
kvsetkeys
etoolbox
xcolor
auxhook
bigintcalc
bitset
etexcmds
gettitlestring
hycolor
hyperref
intcalc
kvdefinekeys
letltxmacro
pdfescape
refcount
rerunfilecheck
stringenc
uniquecounter
zapfding
pdftexcmds
infwarerr
fancyvrb
framed
booktabs
mdwtools
grffile
caption
sourcecodepro
amscls
natbib
```

We can do this with `tlmgr`. If we don't the container installs them at build time which is slow.

PR #155, opened 2021-05-22
Docker update

### LINKED ISSUE #71: Add canlang to the Docker image
Add canlang to the Docker image, a specific release of it, so that we have a stable version of the data sets. When we do this, we should delete https://github.com/UBC-DSCI/introduction-to-datascience/blob/dev/data/create_can_lang_variants.R

ttimbers:
Update to Dockerfile and actions that control its building.
- Only build Docker image when we push to master (makes cleaner image tags)
- Added canlang R package and LaTeX packages
- Added `docker-compose.yml` so that we can launch the RStudio container via `docker-compose up -d` to launch the rstudio web app, and `docker-compose run --rm rstudio bash` to run the container with bash as the entry point. This is less to type, and correctly specifies the ports, volume path and env vars.

Closes #71 and  #84.

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,7 +36,7 @@ This is the source for the Introduction to Data Science textbook.
 > You can use this docker container to edit the files without RStudio using either vim or emacs. This can be done via:
 >
 > ```
-> docker-compose run --rm rstudio bash
+> docker-compose run --rm book-env bash
 > ```
 > 
 > or via:
--- a/README.md
+++ b/README.md
@@ -32,6 +32,12 @@ This is the source for the Introduction to Data Science textbook.
 
 > ### Usage without RStudio
 > You can use this docker container to edit the files without RStudio using either vim or emacs. This can be done via:
+>
+> ```
+> docker-compose run rstudio bash
+> ```
+> 
+> or via:
 > ```
 > docker run --rm -it -v $PWD:/introduction-to-datascience ubcdsci/intro-to-ds /bin/bash
 > ```
--- a/README.md
+++ b/README.md
@@ -10,8 +10,10 @@ This is the source for the Introduction to Data Science textbook.
     - type the following in terminal:
 
     ```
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -24,6 +34,12 @@ This is the source for the Introduction to Data Science textbook.
 
 > ### Usage without RStudio
 > You can use this docker container to edit the files without RStudio using either vim or emacs. This can be done via:
+>
+> ```
+> docker-compose run --rm rstudio bash
```
trevorcampbell, line 39: this shouldn't be in the "using without Rstudio" section, right?

docker-compose.yml
```diff
@@ -0,0 +1,10 @@
+services:
+  rstudio:
+    image: "ubcdsci/intro-to-ds:v0.6.0-47d5cbf"
```
trevorcampbell, line 3: is this docker-compose.yml file auto-generated?

- if so: do we want it to be committed to the repo?
- if not: do we want that `v0.6.0-47blahblah` to be hard-coded?

trevorcampbell (commented):
looks good to me -- i just have a few confusions that are almost certainly because i'm a GH workflows novice.

README.md
ttimbers: It should be, rstudio is just the nickname for the container. But now this makes me think this is not the best nickname... I'll change this to  `book-env`.

docker-compose.yml
ttimbers: b5c491e and f16fd62 should fix this

README.md
ttimbers: 90bf19c and 4184d39 address this.

ttimbers:
OK - still need to fix this! I am currently just printing the regex and not updating the file 🤦‍♀️ !

ttimbers:
fixed here: 72c7358

PR #157, opened 2021-05-23
Docker update

ttimbers:


ttimbers:
Catching up to what I tested out and got working in master

PR #159, opened 2021-05-23
Remove unneeded quote from docker-compose.yml

ttimbers:


ttimbers:
I did this another way, please ignore.

PR #161, opened 2021-05-25
fixed issue with cluster plot

leem44:
@ttimbers I removed x1 and x2 which was preventing the plot from rendering! Thank you for noticing that!


ttimbers:
I just tested this fix and the book now builds in the new Docker image (v0.10.0). Thanks for this and merging!

Issue #153, opened 2021-05-18
Base pipe feature released in R 4.1


ttimbers:
@trevorcampbell @leem44 - a new base pipe feature was just released in R 4.1. I think we have to think about incorporating this in the book. We don't have to make this decision today, but we should think carefully before we go to print. This is something that could easily date our book really fast if the community adopts it and we do not...

- https://twitter.com/thomas_mock/status/1394693648143499269
- https://stat.ethz.ch/R-manual/R-devel/doc/html/NEWS.html (search for pipe)

ttimbers:
I had a very helpful conversation with @jennybc about this, which I will attempt to summarize below:

Jenny said that the Tidyverse has (long) had an [official policy of supporting past R versions](https://www.tidyverse.org/blog/2019/04/r-version-support/), and what that means for them is that they cannot switch completely over to the new pipe for many years. She thinks this means that there will be some awkward period where this has to be addressed in books and documentation (I think the publishing of our book will fall squarely in that time period...).

She also said that luckily, `magrittr` (where the Tidyverse pipe comes from) was recently updated to make it more like the built-in pipe, so that lots of code will work with either. I think that we do need to be cautious with the book if we adopt the built-in pipe as she also said that a built-in pipe can be implemented in a different way than is possible with an add-on package. 

In summary, she recommended (and I wholeheartedly agree) that we should move to the new, forward-looking syntax (the new built-in pipe). However, we should introduce both (because they will see both in the docs, and in any place they search for help online), and then move to just using the new, built-in one. The one other place where we might want to again discuss both are if we have any examples in the book where the  behaviour would differ between the two pipes. There we would be well served to call attention to these differences.

Thanks again @jennybc for your help and advice here!

trevorcampbell:
+1 to switching to the new pipe but introduce both and tell readers about the difference

jennybc:
Blog posts related to magrittr pipe vs base R pipe:

* <https://www.tidyverse.org/blog/2020/08/magrittr-2-0/> planning, asking for feedback
* <https://www.tidyverse.org/blog/2020/11/magrittr-2-0-is-here/> what actually happened

ttimbers:
Thanks for these refs @jennybc!

trevorcampbell:
I believe this is handled now. Closing (but reopen if I'm wrong!)

PR #143, opened 2021-05-09
Ch1 revision

leem44:
@ttimbers @trevorcampbell -- whoever wants to review this, please take a look when you can!
- addressed issues #114 and #130 
- added explanation of data types and structures in wrangling chapter 
- provided explanation of how to read book 
- added data science workflow diagram with chapters 
- added diagrams of the more complicated functions such as `ggplot` and `mutate`
- kept simplest use cases of `select` & `filter` in chapter one and moved others to wrangling (chapter 3) and edited text where appropriate
- moved mutate to wrangling chapter and changed things as needed
- updated learning objective to reflect changes 
- added learning objective about data types to wrangling chapter 
- removed scatterplot example in chapter 1 and changed it to a simpler bar chart example
- overall went through the writing and changed things for clarity
- removed Jupyter content

note:
- chapter numbers and names in the data science workflow diagram may need to be updated to match the new name of the chapter

Issue #131, opened 2021-04-08
Forward references & unexplained concepts


ttimbers:
Go through book and look for forward references or unexplained concepts. **MINOR**

Issue #127, opened 2021-04-08
Update to the newest `dplyr` and `ggplot2`


ttimbers:
Check most recent updates to `dplyr` and `ggplot2` and make sure we are using the most up to date syntax in the books. If this needs to change, also change in the worksheets. This addresses comments made by Rev A & E. **MINOR**

PR #180, opened 2021-06-05
forward references, new `ggplot` `dplyr` function, intended audience

leem44:
- addressing #131, #127, #121, #153, #183
addressed these for wrangling and index chapters here:  #143
- addressed comments here: #102
   - The items I didn't cross out: I either created a separate issue or I decided not to implement because I didn't agree or they didn't require changing (if anyone feels strongly that I should then I am happy to discuss!)


leem44:
@trevorcampbell @ttimbers please have a look at these changes! Thanks!

trevorcampbell:
@leem44 not doing a full review, but I did do a quick check for `%>%` and there are at least a few of them still remaining in the various Rmd files. I was assuming we'd introduce both options when we introduce the pipe, and then just stick with the newer one, no?

leem44:
@trevorcampbell I did leave it in the wrangling chapter as a "here is what the old pipe looks like but we're not using it" type of thing. Did I leave it in any other chapters?

trevorcampbell:
It appears in 01-reading, and 02-wrangling it seems that only `%>%` is introduced, and also appears in 04-version_control. Maybe not editing because some of these are part of other PRs?

leem44:
Yes that's right -- I should have mentioned that I didn't update 04-version_control because of the pending PR

trevorcampbell:
@leem44 two notes:

- there are now conflicts between this branch and `dev` (because I merged #146 ). For most cases we should accept the version that is now in `dev`.
- the old pipe appears in `05-...` through `08-...` (due to the merged PR #146 ) -- when resolving merge conflicts we should keep the new pipe operator

PR #219, opened 2021-07-24
inference chapter specific

### LINKED ISSUE #111: Review: Ch 11 (inference)
Reviewer E: 
- In overview, give an example of the type of question inference asks and clarify how different than prediction
  - **TC: tentatively agree, but not strongly. will think about it**
- I really like the simulation-based approach to inference used here!
- There is another confusing location of an R-generated plot here. Use more proscriptive “In Figure X” to clarify instead of “below”
  - **TC: need to fix this, i see at least one instance of it** update: fixed
- Although not using analytical approach, it might be worth formally mentioning foundational concepts like the central limit theorem and law of large numbers since students are likely to encounter these in any future study of inference
  - TC: nope, too technical
- Great diagrams throughout the chapter to explain key concepts!
- It’s possibly worth mentioning that in some cases closed form solutions do exist even if you don’t wish to discuss. For example, closed form solutions for something like a binomial probability estimation are so straightforward and well-understood, it might be wasteful to bootstrap to figure this out (outside of a learning context)
  - TC: way too technical

trevorcampbell:
This PR resolves #111 and makes some other improvements.

### DIFF
--- a/inference.Rmd
+++ b/inference.Rmd
@@ -25,14 +25,14 @@ min_x <- function(dist) {
 
 
 ## Overview
-A typical data analysis task in practice is to draw conclusions about some unknown
-aspect of a population of interest based on observed data sampled from that
-population; we typically do not get data on the *entire* population.
-Data analysis questions regarding how summaries, 
-patterns, trends, or relationships in a data set 
-extend to the wider population are called *inferential questions*. This chapter will start
-with the fundamental ideas of sampling from populations and then introduce two common techniques in statistical inference: *point estimation* and
-*interval estimation*. 
+A typical data analysis task in practice is to draw conclusions about some
+unknown aspect of a population of interest based on observed data sampled from
+that population; we typically do not get data on the *entire* population.  Data
+analysis questions regarding how summaries, patterns, trends, or relationships
+in a data set extend to the wider population are called *inferential
+questions*. This chapter will start with the fundamental ideas of sampling from
+populations and then introduce two common techniques in statistical inference:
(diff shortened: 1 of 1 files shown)


inference.Rmd
```diff
@@ -431,44 +544,61 @@ grid.arrange(annotated_sampling_dist_20,
...
 > If samples are selected from a finite population as we are doing in this chapter, we should apply a finite population correction. We multiply $\frac{\sigma}{\sqrt{n}}$ by $\sqrt{\frac{N - n}{N - 1}}$ where $N$ is the population size and $n$ is the sample size. If our sample size, $n$, is small relative to the population size, this finite correction factor is less important. 
 ---> 
 
 ### Summary
-1. A *point estimate* is a single value computed using a sample from a population (e.g. a mean or proportion)
-2. The *sampling distribution* of an estimate is the distribution of the estimate for all possible samples of a fixed size from the same population.
-3. The sample means and proportions calculated from samples are centred around the population mean and proportion, respectively.
+1. A point estimate is a single value computed using a sample from a population (e.g. a mean or proportion)
+2. The sampling distribution of an estimate is the distribution of the estimate for all possible samples of a fixed size from the same population.
+3. The shape of the sampling distribution is usually bell-shaped with one peak and centred at the population mean or proportion.
```
leem44, line 563: Perhaps we should add something to this sentence? Some kind of hedging (e.g. “if our random sample is large enough” type of thing)

trevorcampbell: added a note box above it referencing figrue 10.7!

Issue #228, opened 2021-08-07
Annotate 3D plot in classification chapter


leem44:
Punting to next version of book. From Tiffany's comments in #92: 
for 3 features and annotate our 3D plot like we do for our 2D plot.

leem44:
done

Issue #205, opened 2021-07-17
standardized vs nonstandardized data figure in classification -- add an inset


trevorcampbell:
The left half of what is currently Figure 6.10 (that could change) is hard to interpret because all of the neighbour lines lie on top of one another. We might want to add an inset here that zooms in on the relevant area to show the neighbours better.

leem44:
this is done

Issue #184, opened 2021-06-05
explain average and standard deviation


leem44:
We briefly mention these terms in the early chapters and then we start using them in classification 
I think we should explain what an average and standard deviation is. Not sure if we want to give a formula or explain it in broad terms, but students often ask what the standard deviation is when we get to the scaling part. Not sure what chapter would be best to explain it since we mention it in early chapters (reading, wrangling), but I don't think it makes sense to define it precisely there.

Issue #107, opened 2021-03-28
Review: Ch 7 (classification_continued) 


leem44:
# Reviewer E: 
- I love the cross validation diagram -- extremely helpful!
  - *ML: no changes needed here* 
- ~Consider defining “accuracy” more precisely and either defining “Kap” or removing it from tidymodels output. It’s a bit distracting to have it reported but unexplained~
- ~Consider spending slightly more time explaining confusion matrix and what each cell means~
- ~Chapter 7 is extremely dense and hits on so many foundational modeling concepts. I think some of this could be helpful to pull up before Chapter 6 and describe a holistic modeling workflow~

leem44:
Reviewer B: 
- Wow, cross-validation in an intro course! The future has arrived! Cross-validation based error-estimates are much easier to understand than summaries like R-squared or AIC.

trevorcampbell:
# Reviewer D
- it is unclear to me what the overall workflow advocated by authors should be
- Should the readers/users split the data into training set, validation set and test set (such that the training set and the validation set combine to yield the overall training set)?
Should they then tune the classifier by building it on the training set and evaluating it on the validation set? 
Once the classifier is tuned, should they assess its accuracy on the test set?  
Once accuracy of tuned classifier is established, should the classification model/technique be applied to the ENTIRE data (i.e., training set + validation set + test set) to perform classifications for new observations?
- If the 4 items described above capture the workflow intended by the authors, I find it confusing that this workflow is presented backwards in the manuscript – for example, the accuracy of the classifier is assessed BEFORE the classifier is actually tuned.
- Can the authors clarify their intended workflow and make sure the chapters in which they present the elements of this workflow follow a logical sequence?

trevorcampbell:
# Reviewer A
- "precision and recall, not discussed": Why not?
- p163: the figures are hard to follow (the flow of them) because they get jumbled in the PDf version. in the HTML version it's fine. actual comment: I expected to see the scatterplot as the first graphic. If you want the reader to see these in print, be sure to point to them specifically in the text.
- p165 first code block: I think it's important to point out that you are not simply sampling the rows of the cancer data set, but are performing a stratified sample.
- p166 You could emphasize the stratification by summarizing the split between M and B in each data set.
- p169 Is that good? I feel like some wrap up of the performance would be good for the novice reader here.
- 170 Does that mean you pool all the data and train your final classifier? Be very clear for the reader since this is an intro text.
[...]

PR #234, opened 2021-08-16
Review classification chapters 

leem44:
@ttimbers @trevorcampbell PR for classification chapters review! 
addressing issues: #106, #107, #92 and #93  (I either addressed it, added a rebuttal or made a new issue)
Also addressed these: #184, #228, #205

### DIFF
--- a/classification1.Rmd
+++ b/classification1.Rmd
@@ -610,7 +610,7 @@ especially if we want to handle multiple classes, more than two variables,
 or predicting the class for multiple new observations. Thankfully, in R,
 the $K$-nearest neighbors algorithm is implemented in the `parsnip` package 
 included in the 
-[`tidymodels` meta package](https://www.tidymodels.org/), along with 
+[`tidymodels` package](https://www.tidymodels.org/), along with 
 many [other models](https://www.tidymodels.org/find/parsnip/)
  that you will encounter in this and future chapters of the book. The `tidymodels` collection
 provides tools to help make and use models, such as classifiers.  Using the packages
@@ -627,7 +627,7 @@ We will use the `cancer` data set from above, with
 perimeter and concavity as predictors and $K = 5$ neighbors to build our classifier. Then
 we will use the classifier to predict the diagnosis label for a new observation with
 perimeter 0, concavity 3.5, and an unknown diagnosis label. Let's pick out our two desired
-predictor variables and class label and store it as a new dataset named `cancer_train`:
+predictor variables and class label and store them as a new data set named `cancer_train`:
 
 ```{r 05-tidymodels-2}
 cancer_train <- cancer |>
@@ -655,7 +655,7 @@ knn_spec
 ```
 
 In order to fit the model on the breast cancer data, we need to pass the model specification
-and the dataset to the `fit` function. We also need to specify what variables to use as predictors
+and the data setto the `fit` function. We also need to specify what variables to use as predictors
 and what variable to use as the target. Below, the `Class ~ Perimeter + Concavity` argument specifies 
 that `Class` is the target variable (the one we want to predict),
 and both `Perimeter` and `Concavity` are to be used as the predictors.
@@ -698,8 +698,8 @@ predict(knn_fit, new_obs)
 
 Is this predicted malignant label the true class for this observation? 
 Well, we don't know because we do not have this
-observation's diagnosis&mdash; that is what we were trying to predict.
-In the next chapter, we will 
+observation's diagnosis&mdash; that is what we were trying to predict! The 
+classifier's prediction is not necessarily correct, but in the next chapter, we will 
 learn ways to quantify how accurate we think our predictions are.
 
 ## Data preprocessing with `tidymodels`
@@ -731,7 +731,8 @@ $K$-nearest neighbor classification algorithm, this large shift can change the
 outcome of using many other predictive models. 
 
 To scale and center our data, we need to find
-our variables' mean and *standard deviation* (a number quantifying how spread out values are). 
+our variables' *mean* (the average, which quantifies the "central" value of a 
+set of numbers) and *standard deviation* (a number quantifying how spread out values are). 
 For each observed value of the variable, we subtract the mean (center the variable) 
 and divide by the standard deviation (scale the variable). When we do this, the data 
 is said to be *standardized*, and all variables in a data set will have a mean of 0 
@@ -795,7 +796,7 @@ For example:
 You can find [a full set of all the steps and variable selection functions](https://tidymodels.github.io/recipes/reference/index.html)
 on the recipes home page.
 
-Here we have calculated the required statistics based on the data input into the 
+At this point, we have calculated the required statistics based on the data input into the 
 recipe, but the data are not yet scaled and centred. To actually scale and center 
 the data, we need to apply the bake function to the unscaled data.
 
@@ -805,10 +806,10 @@ scaled_cancer
 ```
 
 It may seem redundant that we had to both `bake` *and* `prep` to scale and center the data.
-However, we do this in two steps so we could specify a different data set in the `bake` step
-if desired, say, new data you want to predict, which were not part of the training set. 
+ However, we do this in two steps so we can specify a different data set in the `bake` step, 
+ for instance, new data that were not part of the training set. 
 
-At this point, you may wonder why we are doing so much work just to center and
+You may wonder why we are doing so much work just to center and
 scale our variables.  Can't we just manually scale and center the `Area` and
 `Smoothness` variables ourselves before building our $K$-nearest neighbor model? Well,
 technically *yes*; but doing so is error-prone.  In particular, we might
@@ -951,7 +952,10 @@ ggplot(unscaled_cancer, aes(x = Area, y = Smoothness, group = Class, color = Cla
     xend = unlist(neighbors[3, attrs[1]]),
     yend = unlist(neighbors[3, attrs[2]])
   ), color = "black") +   theme_light() +  
- facet_zoom( xlim= c(399.7, 401.6), ylim = c(0.08, 0.14), zoom.size = 2) 
+# facet_zoom( xlim = c(399.7, 401.6), ylim = c(0.08, 0.14), zoom.size = 2) + 
+   facet_zoom(x = ( Area > 380 & Area < 420) , 
+              y = (Smoothness > 0.08 & Smoothness < 0.14), zoom.size = 2) + 
+  theme_bw()
 ```
 
 ### Balancing
@@ -1000,7 +1004,10 @@ rare_plot
 > process, which then guarantees the same result, i.e., the same choice of 3
 > observations, each time the code is run. In general, when your code involves
 > random numbers, if you want *the same result* each time, you should use
-> `set.seed`; if you want a *different result* each time, you should not.
+> `set.seed`; if you want a *different result* each time, you should not. 
+> You only need to `set.seed` once at the beginning of your analysis, so the 
+rest of the analysis uses seemingly random numbers.
+ 
 
 Suppose we now decided to use $K = 7$ in $K$-nearest neighbor classification.
 With only 3 observations of malignant tumors, the classifier 
--- a/classification1.Rmd
+++ b/classification1.Rmd
@@ -84,7 +84,7 @@ the diagnosing physician is. Furthermore, benign tumours are not normally
 dangerous; the cells stay in the same place and the tumour stops growing before
 it gets very large. By contrast, in malignant tumours, the cells invade the
 surrounding tissue and spread into nearby organs where they can cause serious
-damage ([learn more about cancer here](https://www.worldwidecancerresearch.org/who-we-are/cancer-basics/)).
+damage (@stanfordhealthcare).
 Thus, it is important to quickly and accurately diagnose the tumour type to
 guide patient treatment.
 
@@ -738,9 +738,8 @@ is only one black line on this plot).  Here the computation of nearest
 neighbours is dominated by the much larger-scale area variable. On the right,
 the plot for standardized data shows a much more intuitively reasonable
 selection of nearest neighbours. Thus, standardizing the data can change things
-in an important way when we are using predictive algorithms.  As a rule of
-thumb, standardizing your data should be a part of the preprocessing you do
-before any predictive modelling / analysis.
+in an important way when we are using predictive algorithms.  Standardizing your data should be a part of the preprocessing you do
+before predictive modelling / analysis, and its important to think about whether the problem domain and whether you need to standardize your data.
 
 ```{r 05-scaling-plt, echo = FALSE, fig.height = 4, fig.width = 10, fig.cap = "Comparison of K = 3 nearest neighbours with standardized and unstandardized data"}
 
--- a/classification1.Rmd
+++ b/classification1.Rmd
@@ -52,8 +52,8 @@ we use these data to train, or teach, our classifier. Once taught, we can use
 the classifier to make predictions on new data for which we do not know the class.
 
 There are many possible methods that we could use to predict
-a categorical class/label for an observation. In this book we will
-focus on the simple, widely used **K-nearest neighbors** algorithm.
+a categorical class/label for an observation. In this book, we will
+focus on the widely used **K-nearest neighbors** algorithm.
 In your future studies, you might encounter decision trees, support vector machines (SVMs),
 logistic regression, neural networks, and more; see the additional resources
 section at the end of the next chapter for where to begin learning more about
@@ -70,22 +70,22 @@ In this chapter and the next, we will study a data set of
 [digitized breast cancer image features](http://archive.ics.uci.edu/ml/datasets/Breast+Cancer+Wisconsin+%28Diagnostic%29),
 created by Dr. William H. Wolberg, W. Nick Street, and Olvi L. Mangasarian at
 the University of Wisconsin, Madison. Each row in the data set represents an
-image of a tumour sample including the diagnosis (benign or malignant) and
+image of a tumor sample, including the diagnosis (benign or malignant) and
 several other measurements (nucleus texture, perimeter, area, and more).
 Diagnosis for each image was conducted by physicians. 
 
 As with all data analyses, we first need to formulate a precise question that
-we want to answer. Here, the question is *predictive*: can we use the tumour
-image measurements available to us to predict whether a future tumour image
-(with unknown diagnosis) shows a benign or malignant tumour? Answering this
-question is important because traditional, non-data-driven methods for tumour
+we want to answer. Here, the question is *predictive*: can we use the tumor
+image measurements available to us to predict whether a future tumor image
+(with unknown diagnosis) shows a benign or malignant tumor? Answering this
+question is important because traditional, non-data-driven methods for tumor
 diagnosis are quite subjective and dependent upon how skilled and experienced
-the diagnosing physician is. Furthermore, benign tumours are not normally
-dangerous; the cells stay in the same place and the tumour stops growing before
-it gets very large. By contrast, in malignant tumours, the cells invade the
+the diagnosing physician is. Furthermore, benign tumors are not normally
+dangerous; the cells stay in the same place and the tumor stops growing before
+it gets very large. By contrast, in malignant tumors, the cells invade the
 surrounding tissue and spread into nearby organs where they can cause serious
 damage (@stanfordhealthcare).
-Thus, it is important to quickly and accurately diagnose the tumour type to
+Thus, it is important to quickly and accurately diagnose the tumor type to
 guide patient treatment.
 
 ### Loading the cancer data
@@ -109,7 +109,7 @@ cancer
 
 ### Describing the variables in the cancer data set
 
-Breast tumours can be diagnosed by performing a *biopsy*, a process where
+Breast tumors can be diagnosed by performing a *biopsy*, a process where
 tissue is removed from the body and examined for the presence of disease.
 Traditionally these procedures were quite invasive; modern methods such as fine
 needle asipiration, used to collect the present data set, extract only a small
@@ -175,9 +175,9 @@ cancer |>
 
 Before we start doing any modelling, let's explore our data set. Below we use
 the `group_by`, `summarize` and `n` functions to find the number and percentage 
-of benign and maligant tumour observations in our data set. The `n` function within
+of benign and maligant tumor observations in our data set. The `n` function within
 the `summarize` function counts the number of observations in each `Class` group. 
-We have 357 (63\%) benign and 212 (37\%) malignant tumour observations.
+We have 357 (63\%) benign and 212 (37\%) malignant tumor observations.
 
 ```{r 05-tally}
 num_obs <- nrow(cancer)
@@ -213,7 +213,7 @@ tend to have lower concavity and perimeter values, and malignant ones tends to h
 concavity and perimeter values. Suppose we
 obtain a new observation not in the current data set that has all the variables
 measured *except* the label (i.e., an image without the physician's diagnosis
-for the tumour class). We could compute the perimeter and concavity values,
+for the tumor class). We could compute the perimeter and concavity values,
 resulting in values of, say, 1 and 1. Could we use this information to classify
 that observation as benign or malignant? Based on the scatter plot, how might 
 you classify that new observation? How would you classify a new observation with
@@ -544,8 +544,7 @@ scaled_cancer_3 |>
     xaxis = list(title = "Perimeter (standardized)"),
     yaxis = list(title = "Concavity (standardized)"),
     zaxis = list(title = "Symmetry (standardized)")
-  )) 
-
+  ))
 ```
 
 *Click and drag the plot above to rotate it, and scroll to zoom. Note that in
@@ -913,7 +912,7 @@ detection, there are many cases in which the "important" class to identify
 class (no disease, normal email).
 
 To better illustrate the problem, let's revisit the scaled breast cancer data, 
-`cancer`; except now we will remove many of the observations of malignant tumours, simulating
+`cancer`; except now we will remove many of the observations of malignant tumors, simulating
 what the data would look like if the cancer was rare. We will do this by
 picking only 3 observations randomly from the malignant group, and keeping all
 of the benign observations. We choose these 3 observations using the `slice_sample`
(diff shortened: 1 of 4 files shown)


classification1.Rmd
```diff
@@ -560,17 +646,17 @@ cancer_train <- cancer |>
...
-which weight each neighbour's  vote differently,  can be found on 
+how neighbors vote when classifying a new observation; by setting it to `"rectangular"`,
+each of the $K$ nearest neighbors gets exactly 1 vote as described above. Other choices, 
+which weigh each neighbor's vote differently, can be found on 
 [the tidymodels website](https://parsnip.tidymodels.org/reference/nearest_neighbor.html).
-We specify the particular computational
-engine (in this case, the `kknn` engine) for training the model with the `set_engine` function.
-Finally we specify that this is a classification problem with the `set_mode` function.
+In the `set_engine` argument, we specify which package or system will be used for training
+the model. In this case, `kknn` is an R package for performing $K$-nearest neighbors classification.
```
trevorcampbell, line 658: "In this case, kknn is an R package..." -- sentence a bit awk flow from previous

classification1.Rmd
```diff
@@ -602,16 +688,16 @@ knn_fit
...
 Here you can see the final trained model summary. It confirms that the computational engine used
 to train the model  was `kknn::train.kknn`. It also shows the fraction of errors made by
-the nearest neighbour model, but we will ignore this for now and discuss it in more detail
+the nearest neighbor model, but we will ignore this for now and discuss it in more detail
 in the next chapter.
 Finally it shows (somewhat confusingly) that the "best" weight function 
 was "rectangular" and "best" setting of $K$ was 5; but since we specified these earlier,
 R is just repeating those settings to us here. In the next chapter, we will actually
-let R tune the model for us. 
+let R find the $K$ value for us.
```
trevorcampbell, line 696: value of K

classification2.Rmd
```diff
@@ -1012,7 +1050,7 @@ one over increasing predictor set sizes
 and another to check which predictor to add in each round (where you see `for (j in 1:length(names))` below).
 For each set of predictors to try, we construct a model formula,
 pass it into a `recipe`, build a `workflow` that tunes
-a KNN classifier using 5-fold cross-validation, 
+a $K$-NN classifier using 5-fold cross-validation,
```
trevorcampbell, line 1053: I thought I saw that you were changing to K-nearest neighbors everywhere. I might have missed something though, so take with a grain of salt.

references.bib
```diff
@@ -218,3 +219,10 @@ @inproceedings{kluyver2016jupyter
   volume = {87},
   address = {Amsterdam}
 }
+
+@online{stanfordhealthcare,
```
trevorcampbell, line 223: There's something weird about how this is rendering. I see this on the rendered page:

"Stanford Health Care. 2021. “What Is Cancer?” 2021. https://stanfordhealthcare.org/medical-conditions/cancer/cancer.html."

2021 is printed twice...

classification1.Rmd
```diff
@@ -206,28 +200,28 @@ changing them to "Benign" and "Malignant" using the `labels` argument.
...
 you classify that new observation? How would you classify a new observation with
-perimeter value of -1 and concavity value of -0.5? What about 0 and 1? It seems
-like the *prediction of an unobserved label* might be possible, based on our
-visualization. 
+a standardized perimeter value of -1 and a concavity value of -0.5? What about 0 and 1? 
+Based on our visualization, it seems
+like the *prediction of an unobserved label* might be possible.
 
-## Classification with K-nearest neighbours
+## Classification with $K$-nearest neighbors
```
trevorcampbell, line 224: I agree with this change in general. But I have noticed in the HTML navigation table that appears on the left, the $K$ symbol is really, really small.

I am not sure this is an issue we should fix, since it seems to be a problem with how rbookdown renders math in the navigation pane.

classification1.Rmd
```diff
@@ -163,13 +157,14 @@ glimpse(cancer)
...
+can verify the levels of the `Class` column by using the `levels` function.
 This function should return the name of each category in that column. Given
-that we only have 2 different values in our `Class` column (B and M), we
-only expect to get two names back.  Note that the `levels` function requires
-a *vector* argument, while the `select` function outputs a *data frame*; 
-so we use the `pull` function, which converts a single
-column of a data frame into a vector.
+that we only have two different values in our `Class` column (B for benign and M 
+for malignant), we only expect to get two names back.  Note that the `levels` function requires
+a *vector* argument (while the `select` function outputs a *data frame*);
```
trevorcampbell, line 164: why does this part mention the `select` function? This must be based on some old code we had here...

classification1.Rmd
```diff
@@ -163,13 +157,14 @@ glimpse(cancer)
...
-column of a data frame into a vector.
+that we only have two different values in our `Class` column (B for benign and M 
+for malignant), we only expect to get two names back.  Note that the `levels` function requires
+a *vector* argument (while the `select` function outputs a *data frame*); 
+so we use the `pull` function to convert the `Class`
+column into a vector and pass that into the `levels` function to see the categories 
+in the `Class` column. 
 
 ```{r 05-levels}
 cancer |>
```
trevorcampbell, line 170: "turns data frame into a vector"

this is not quite right, and a bit confusing --a dataframe is secretly a list of vectors, and `pull` allows you to actually take a vector out (`select` with one column would still be a list of vectors)

of course my explanation there is way too complicated, but we need some more careful version of that

classification1.Rmd
```diff
@@ -206,28 +200,28 @@ changing them to "Benign" and "Malignant" using the `labels` argument.
...
-concavity and perimeter values. Suppose we
+the upper right-hand corner of the plot area. By contrast, benign
+observations typically fall in the lower left-hand corner of the plot. In other words,
+benign observations tend to have lower concavity and perimeter values, and malignant
+ones tend to have larger values. Suppose we
 obtain a new observation not in the current data set that has all the variables
 measured *except* the label (i.e., an image without the physician's diagnosis
-for the tumour class). We could compute the perimeter and concavity values,
+for the tumor class). We could compute the standardized perimeter and concavity values,
 resulting in values of, say, 1 and 1. Could we use this information to classify
```
trevorcampbell, line 217: I think we might want to suggest a conclusion for each of these. For example, you could say just purely intuitively "well if the values are 1 and 1, the point would lie in the middle of the orange cloud. it's probably orange then"

trevorcampbell: and maybe reduce the number of these

classification1.Rmd
```diff
@@ -256,115 +250,129 @@ table_with_distances <- function(training, new_point) {
...
 they would have the same diagnosis. 
 
 
-```{r 05-knn-2, echo = FALSE, fig.height = 4, fig.width = 5, fig.cap="Scatter plot of concavity versus perimeter, with malignant nearest neighbour to a new observation highlighted"}
-perim_concav + geom_point(aes(x = new_point[1], y = new_point[2]),
-  color = "red",
-  size = 2.5, 
-  pch = 17
-) +
+```{r 05-knn-2, echo = FALSE, fig.height = 4, fig.width = 5, fig.cap="Scatter plot of concavity versus perimeter, with malignant nearest neighbor to a new observation highlighted"}
```
trevorcampbell, line 298: "with malignant nearest neighbor to a new observation highlighted"

bit awk.

classification1.Rmd
```diff
@@ -425,20 +442,20 @@ cancer |>
 ```
```
trevorcampbell, line 442: right above this line

"subset the first 5 rows"

I'm not a fan of using "subset" as a verb

classification1.Rmd
```diff
@@ -472,86 +489,155 @@ perim_concav + annotate("path",
...
 Although the above description is directed toward two predictor variables, 
-exactly the same K-nearest neighbours algorithm applies when you
+exactly the same $K$-nearest neighbors algorithm applies when you
 have a higher number of predictor variables (i.e., a higher-dimensional
 predictor space).  Each predictor variable may give us new
 information to help create our classifier.  The only difference is the formula
-for the distance between points. In particular, let's say we have $m$ predictor
+for the distance between points. 
+
+Suppose we have $m$ predictor
```
trevorcampbell, line 498: the spacing of these few lines is weird. this one sentence shouldn't be its own paragraph

classification1.Rmd
```diff
@@ -472,86 +489,155 @@ perim_concav + annotate("path",
 ### More than two explanatory variables 
 
 Although the above description is directed toward two predictor variables, 
-exactly the same K-nearest neighbours algorithm applies when you
+exactly the same $K$-nearest neighbors algorithm applies when you
 have a higher number of predictor variables (i.e., a higher-dimensional
```
trevorcampbell, line 493: (i.e. a higher-dimensional predictor space)

do we ever use the word "dimension" or "space" anywhere else in the book? If not, should remove it in this little blurb here.

If so, should consider whether there is simpler language to use here.

classification1.Rmd
```diff
@@ -472,86 +489,155 @@ perim_concav + annotate("path",
...
 
 This formula still corresponds to a straight-line distance, just in a space with 
-more dimensions. For example, suppose we decide to use 3 predictor variables (so, a 3-dimensional space):
-perimeter, concavity, and symmetry. Figure \@ref(fig:05-more) shows what
-the data look like when we visualize them as a 3-dimensional scatter.
+more dimensions. Suppose we want to calculate the distance between a new observation 
+that we want to classify with a perimeter of 0, concavity of 3.5 and symmetry of 1 and
+another observation with a perimeter, concavity and symmetry of 0.417, 2.31 and 0.837 respectively. 
+We have three predictor variables, perimeter, concavity, and symmetry (so, a 3-dimensional space) for two observations where
+$\mathrm{observation}_{old} = (0.417, 2.31, 0.837)$ and $\mathrm{observation}_{new} = (0, 3.5, 1)$.
```
trevorcampbell, line 512: this latex needs some work

in fact, I would just rewrite this to entirely get rid of \mathrm{observation}_{old}. No need to make named observations here

classification1.Rmd
trevorcampbell: I see some instances of this below too, FYI

classification1.Rmd
```diff
@@ -472,86 +489,155 @@ perim_concav + annotate("path",
...
+Based on $K=5$ nearest neighbors with these three predictors we would classify the new observation as malignant since 4 out of 5 of the nearest neighbors are malignant class. 
+Figure \@ref(fig:05-more) shows what the data look like when we visualize them 
+as a 3-dimensional scatter with lines from the new observation to its five nearest neighbors.
 
 ```{r 05-more, echo = FALSE, message = FALSE, fig.cap = "3D scatter plot of the symmetry, concavity, and perimeter variables."}
 library(plotly)
-cancer |>
-  plot_ly(
-    x = ~Perimeter,
+attrs <- c("Perimeter", "Concavity", "Symmetry")
```
trevorcampbell, line 539: this scatter is so incredibly cool. well done.

you should make an issue to remind us to somehow deal with this for the PDF output though (since we obviously can't do interactive 3D, so need to modify both the figure and the text after it)

classification1.Rmd
```diff
@@ -472,86 +489,155 @@ perim_concav + annotate("path",
...
+## $K$-nearest neighbors with `tidymodels`
 
-Coding the K-nearest neighbours algorithm in R ourselves can get complicated,
+Coding the $K$-nearest neighbors algorithm in R ourselves can get complicated,
 especially if we want to handle multiple classes, more than two variables,
 or predicting the class for multiple new observations. Thankfully, in R,
-the K-nearest neighbours algorithm is implemented in the `parsnip` package 
+the $K$-nearest neighbors algorithm is implemented in the `parsnip` package 
 included in the 
 [`tidymodels` meta package](https://www.tidymodels.org/), along with
```
trevorcampbell, line 624: do we use the term "meta package' anywhere else?

in fact I think when we introduce tidyverse, we specifically just call it a package

classification1.Rmd
```diff
@@ -472,86 +489,155 @@ perim_concav + annotate("path",
...
-Let's walk through how to use `tidymodels` to perform K-nearest neighbours classification. 
-We will use the cancer data set from above, with
-perimeter and concavity as predictors and $K = 5$ neighbours to build our classifier. Then
+Let's walk through how to use `tidymodels` to perform $K$-nearest neighbors classification. 
+We will use the `cancer` data set from above, with
+perimeter and concavity as predictors and $K = 5$ neighbors to build our classifier. Then
 we will use the classifier to predict the diagnosis label for a new observation with
-perimeter 0, concavity 3.5, and an unknown diagnosis label. Let's pick out our 2 desired
+perimeter 0, concavity 3.5, and an unknown diagnosis label. Let's pick out our two desired
 predictor variables and class label and store it as a new dataset named `cancer_train`:
```
trevorcampbell, line 641: "store it" -> "store them"

classification1.Rmd
```diff
@@ -621,19 +707,25 @@ new_obs <- tibble(Perimeter = 0, Concavity = 3.5)
 predict(knn_fit, new_obs)
 ```
 
+Is this predicted malignant label the true class for this observation?
```
trevorcampbell, line 710: this is kind of abrupt / too short -- it's not clear that the point you're trying to get across is that the classifier's prediction is not necessarily true or correct

classification1.Rmd
```diff
@@ -642,17 +734,20 @@ In many other predictive models, the *center* of each variable (e.g., its mean)
...
-K-nearest neighbour classification algorithm, this large shift can change the
+$K$-nearest neighbor classification algorithm, this large shift can change the
 outcome of using many other predictive models. 
 
-When all variables in a data set have a mean (center) of 0
-and a standard deviation (scale) of 1, we say that the data have been
-*standardized*. To illustrate the effect that standardization can have on the K-nearest
-neighbour algorithm, we will read in the original, unscaled Wisconsin breast
+To scale and center our data, we need to find
+our variables' mean and *standard deviation* (a number quantifying how spread out values are).
```
trevorcampbell, line 745: why isn't mean italicized / explained here?

classification1.Rmd
```diff
@@ -709,15 +805,23 @@ For example:
 
 You can find [a full set of all the steps and variable selection functions](https://tidymodels.github.io/recipes/reference/index.html)
 on the recipes home page.
-We finally use the `bake` function to apply the recipe to the unscaled data.
+
+Here we have calculated the required statistics based on the data input into the
```
trevorcampbell, line 809: "Here we have" -> "At this point, we have"

classification1.Rmd
```diff
@@ -709,15 +805,23 @@ For example:
...
+recipe, but the data are not yet scaled and centred. To actually scale and center 
+the data, we need to apply the bake function to the unscaled data.
+
 ```{r 05-scaling-4}
 scaled_cancer <- bake(uc_recipe, unscaled_cancer)
 scaled_cancer
 ```
 
+It may seem redundant that we had to both `bake` *and* `prep` to scale and center the data.
+However, we do this in two steps so we could specify a different data set in the `bake` step
```
trevorcampbell, line 819: "However, we do this..." --- bit of a run-on sentence + misplaced commas.

classification1.Rmd
```diff
@@ -709,15 +805,23 @@ For example:
...
+if desired, say, new data you want to predict, which were not part of the training set. 
+
 At this point, you may wonder why we are doing so much work just to center and
```
trevorcampbell, line 822: "At this point, you may..." -> "You may..."

(since we added "at this point" above, remove it here)

classification1.Rmd
```diff
@@ -729,102 +833,143 @@ yourself. You will see further on in Section
...
-  ), color = "black")
+    xend = unlist(neighbors_scaled[3, attrs[1]]),
+    yend = unlist(neighbors_scaled[3, attrs[2]])
+  ), color = "black", size = 0.5)
 
 gridExtra::grid.arrange(unscaled, scaled, ncol = 2)
 ```
 </center>
 
+```{r 05-scaling-plt-zoomed, fig.height = 4, fig.width = 10, echo = FALSE, fig.cap = "Close up of three nearest neighbors for unstandardized data"}
```
trevorcampbell, line 936: This is a super cool plot. The only comment I have here is that the grey highlight showing where the "zoomed in" region is is a bit unclear.

classification1.Rmd
```diff
@@ -846,71 +991,88 @@ The new imbalanced data is shown in Figure \@ref(fig:05-unbalanced).
...
 > Note: You will see in the code above that we use the `set.seed` function.
 > This is because we are using `slice_sample` to artificially pick only 3 of
-> the malignant tumour observations, which uses random sampling to choose which
+> the malignant tumor observations, which uses random sampling to choose which
 > rows will be in the training set. In order to make the code reproducible, we
 > use `set.seed` to specify where the random number generator starts for this
 > process, which then guarantees the same result, i.e., the same choice of 3
 > observations, each time the code is run. In general, when your code involves
 > random numbers, if you want *the same result* each time, you should use
 > `set.seed`; if you want a *different result* each time, you should not.
```
trevorcampbell, line 1014: maybe add a reminder to this note that in your own data analysis, you only set.seed once at the top of the analysis (so that the rest of the analysis uses numbers that seem to be random)

classification2.Rmd
```diff
@@ -52,23 +52,28 @@ the observations in the test set? One way we can do this is to calculate the
 **prediction accuracy**. This is the fraction of examples for which the
 classifier made the correct prediction. To calculate this we divide the number
 of correct predictions by the number of predictions made. 
-This process is illustrated in Figure \@ref(fig:06-ML-paradigm-test).
-Note that there are other measures for how well classifiers perform, such as *precision* and *recall*; 
+
+$$prediction \; accuracy = \frac{number \; of  \; correct  \; predictions}{total \;  number \;  of  \; predictions}$$
```
trevorcampbell, line 56: OK to use full words here, but use \mathrm to make these things non-italics.

italics in latex is generally reserved for symbols

classification2.Rmd
```diff
@@ -52,23 +52,28 @@ the observations in the test set? One way we can do this is to calculate the
...
-work through an example of this process using the breast cancer dataset.
-We start by loading the necessary libraries, reading in the breast cancer data
+work through an example of this process using the breast cancer data set.
+We start by loading the necessary packages, reading in the breast cancer data
 from the previous chapter, and making a quick scatter plot visualization of
-tumour cell concavity versus smoothness coloured by diagnosis in Figure \@ref(fig:06-precode).
+tumor cell concavity versus smoothness colored by diagnosis in Figure \@ref(fig:06-precode).
 
-```{r 06-precode, fig.height = 4, fig.width = 5, fig.cap="Scatterplot of tumour cell concavity versus smoothness coloured by diagnosis label", message = F, warning = F}
+```{r 06-precode, fig.height = 4, fig.width = 5, fig.cap="Scatterplot of tumor cell concavity versus smoothness coloured by diagnosis label", message = F, warning = F}
```
trevorcampbell, line 76: scatterplot -> scatter plot 

(I believe that is what is more consistently used throughout)

trevorcampbell: also just in general i prefer scatter plot

classification2.Rmd
```diff
@@ -100,29 +105,31 @@ using a larger test data set). Here, we will use 75% of the data for training,
...
 The `training` and `testing` functions then extract the training and testing
 data sets into two separate data frames.
+
 ```{r 06-initial-split}
 set.seed(1)
 cancer_split <- initial_split(cancer, prop = 0.75, strata = Class)
 cancer_train <- training(cancer_split)
-cancer_test <- testing(cancer_split)
+cancer_test <- testing(cancer_split) 
 ```
```
trevorcampbell, line 134: should emphasize again that in a typical data analysis, you set.seed once at the beginning.

trevorcampbell: (below -- github didn't let me comment the relevant line)

classification2.Rmd
```diff
@@ -142,10 +149,31 @@ a train / test split of 75% / 25%, as desired. Recall from Chapter \@ref(classif
...
+
 ### Preprocess the data
 
-As we mentioned last chapter, KNN is sensitive to the scale of the predictors,
-and so we should perform some preprocessing to standardize them. An
+As we mentioned in the last chapter, $K$-nearest neighbors is sensitive to the scale of the predictors,
+so we should perform some preprocessing to standardize them. An
 additional consideration we need to take when doing this is that we should
 create the standardization preprocessor using **only the training data**. This ensures that
 our test data does not influence any aspect of our model training. Once we have
```
trevorcampbell, line 179: below here (github didn't let me comment)

"makes it simple to handle this properly"

I think this was likely something I added -- probably should avoid "simple" in case this is tough for some readers!

classification2.Rmd
```diff
@@ -186,15 +214,15 @@ knn_fit <- workflow() |>
 knn_fit
 ```
 
-> Note: Here again you see the `set.seed` function. In the K-nearest neighbours algorithm, 
-> there is a tie for the majority neighbour class, the winner is randomly selected. Although there is no chance
+> Note: Here again you see the `set.seed` function. In the $K$-nearest neighbors algorithm, 
+> if there is a tie for the majority neighbor class, the winner is randomly selected. Although there is no chance
 > of a tie when $K$ is odd (here $K=3$), it is possible that the code may be changed in the future to have an even value of $K$. 
 > Thus, to prevent potential issues with reproducibility, we have set the seed. Note that in your own code,
```
trevorcampbell, line 220: this note block is quite odd. I'm not sure what it's saying

also "you only have to" -> "you should only" -- we should be very clear that in regular data analyses you set.seed once at the beginning, but for our textbook we've used it more to help us carefully create examples for learning purposes

classification2.Rmd
```diff
@@ -472,8 +500,8 @@ Then, when we create our data analysis workflow, we use the `fit_resamples` func
 instead of the `fit` function for training. This runs cross-validation on each
 train/validation split. 
 
-> **Note:** we set the seed when we call `train` not only because of the potential for ties, but also because we are doing 
-> cross-validation. Cross-validation uses a random process to select how to partition the training data.
+> **Note:** we set the seed because we are doing 
+> cross-validation, which uses a random process to select how to partition the training data.
```
trevorcampbell, line 504: again, should be more careful about this note box

classification2.Rmd
```diff
@@ -527,23 +556,24 @@ vfold_metrics <- workflow() |>
...
 depends entirely on the downstream application of the data analysis. In the
-present situation, we are trying to predict a tumour diagnosis, with expensive,
+present situation, we are trying to predict a tumor diagnosis, with expensive,
 damaging chemo/radiation therapy or patient death as potential consequences of
 misprediction. Hence, we might like to 
 do better than `r round(100*(vfold_metrics %>% filter(.metric == "accuracy"))$mean,0)`% for this application.  
 
 In order to improve our classifier, we have one choice of parameter: the number of
-neighbours, $K$. Since cross-validation helps us evaluate the accuracy of our
+neighbors, $K$. Since cross-validation helps us evaluate the accuracy of our
```
trevorcampbell, line 576: below here 

"we set the seed prior to tuning to ensure results are reproducible"

same comment as above

maybe here just remove this sentence.

classification2.Rmd
```diff
@@ -909,7 +946,7 @@ is less obvious, as all seem like reasonable candidates. It
 is not clear which subset of them will create the best classifier. One could use visualizations and
 other exploratory analyses to try to help understand which variables are potentially relevant, but
 this process is both time-consuming and error-prone when there are many variables to consider.
-We therefore need a more systematic and programmatic way of choosing variables. 
+We therefore, need a more systematic and programmatic way of choosing variables.
```
trevorcampbell, line 949: grammar -- misplaced comma

classification2.Rmd
```diff
@@ -925,7 +962,7 @@ In particular, you
 2. tune each one using cross validation
 3. pick the subset of predictors that gives you the highest cross-validation accuracy  
 
-Best subset selection is applicable to any  classification method (KNN or otherwise).
+Best subset selection is applicable to any classification method ($K$-NN or otherwise).
```
trevorcampbell, line 965: full name? forget if we are doing full name consistently throughout

classification2.Rmd
```diff
@@ -1012,7 +1050,7 @@ one over increasing predictor set sizes
 and another to check which predictor to add in each round (where you see `for (j in 1:length(names))` below).
 For each set of predictors to try, we construct a model formula,
 pass it into a `recipe`, build a `workflow` that tunes
-a KNN classifier using 5-fold cross-validation, 
+a $K$-NN classifier using 5-fold cross-validation, 
 and finally records the estimated accuracy.
 
 ```{r 06-fwdsel-2, warning = FALSE}
```
trevorcampbell, line 1056: This code block has a for loop in it. Tiffany said she would make a package for this, and we should definitely make sure to do that. Might want to ping her to do that and then edit this PR accordingly.

classification1.Rmd
leem44: referenced in new issue #240

classification1.Rmd
leem44: I think we can remove that comment since we provide an explanation of the function in the paragraph above

classification1.Rmd
leem44: Made a new issue #241

classification1.Rmd
leem44: I searched and we use "metapackage" or "meta package" in:
- jupyter.Rmd
- viz.Rmd
- wrangling.Rmd
Should we remove all instances of it? If so, I will make another issue.

trevorcampbell: @leem44 no -- if we use it in a bunch of places it's fine to just leave it as-is (for this PR, anyway)

classification2.Rmd
leem44: I think we have $K$-NN in a couple of places

classification2.Rmd
leem44: #242

classification2.Rmd
trevorcampbell: OK. I don't feel strongly, so happy to keep it

references.bib
leem44: This seems to be an issue with a couple of our references so I made a new issue #243

classification2.Rmd
leem44: as per Trevor's comment this is going to be addressed in global set.seed issue

Issue #160, opened 2021-05-25
Additional functions to add to wrangling?


ttimbers:
We use `arrange` a bit in the worksheets, but I don't think it's currently in the book? We might want to at least add this one to wrangling?

Issue #103, opened 2021-03-28
Review: Ch 3 (wrangling) 


leem44:
# Reviewer E: 
- Chapter 3 could benefit by making sure it is encouraging students to use dplyr in the most effective ways.
   - In Section 3.6, I would have liked to see the scoped verbs (e.g. select_at, summarize_if) 
       - *ML: scoped verbs have been superceded in favor of `across()` https://dplyr.tidyverse.org/reference/summarise_all.html*
   - ~select helpers (e.g. matches, starts_with) covered since this can make coding much more efficient.~ 
   - In fact, summarize_all(), now meets the need that is used to motivate the introduction of the purrr package. More broadly, Section 3.7 should be evaluated to make sure it is using current best practices following the release of dplyr 1.0.0. dplyr was enhanced with the goal of making purrr less necessary when working with tabular data. In addition to the summarize_all() option that I mentioned above, mutate() now vectorizes many functions so one can write mutate(data, x = f(w)) instead of mutate(data, x = map(w, f)). In my experience, purrr seems to confuse newer users, so it may be worth striving to cut this section and using the new dplyr approaches instead. 
     - *ML: mentioned in issue #193
- ~Ch titles/subtitles: Section 3.6-3.7: Both of these use “iterate” in the title to describe the group_by+summarize workflow as well as the purrr workflow. This may lead to confusion on how the two methods are different. I would tend to call group_by+summarize something like “aggregation” or “subgroup analysis/summarization” to distinguish from mapping. I think iteration is a better description for purrr since it works more like traditional for-loop style iteration (same number of elements returned as inputted.)~
- In 3.4:
  - Very nice spiraling of new concepts in this chapter and repeatedly building off of previous concepts! (For example, how the tidyr::seperate() example reuses tidyr::pivot_longer() )
     - *ML: no changes needed here*
  - More weird spacing of figures and text 
     - *ML: mentioned in #96*
  - ~Explicitly try to define “wide” versus “long” before using. It should be intuitive but might trip students up.~ 
  - ~Explain why each situation might arise. Students may wonder why anyone would produce the “wrong” form of data, but discussion could explain, for example, that human-readable tables are often in “wide” format, but for visualization and analysis we want them “long”.~
  - ~For the table that is “too long”, the important column (containing the future column names) is truncated. This makes it harder to understand the concept and why this is a bad format~
[...]

PR #235, opened 2021-08-17
Wrangling chapter review

leem44:
@trevorcampbell @ttimbers Please have a look at this PR and let me know if I should make any changes! 
This PR addresses: #103, #160

- Note: I have *not* yet removed `map` from this chapter  -- I figure its easier to get any feedback on this, make changes and then create a branch off of this to remove `map`


ttimbers:
@leem44 - I combined row and observation figure into one, as the explanation seemed to long and drawn out. Here's what the combined figure looks like:

Screen Shot 2021-09-19 at 12 07 37 PM

Is this change OK?

ttimbers:
Also changed the vector figure to be of type character, as previously it was called an integer, but then we actually made a double vector in code... I tried to swap it to double, but then it was too confusing for a first example. Character vector is a simpler and thus better first example:
Screen Shot 2021-09-19 at 12 23 34 PM

ttimbers:
Also changed the vector figure to be of type character, as previously it was called an integer, but then we actually made a double vector in code... I tried to swap it to double, but then it was too confusing for a first example. Character vector is a simpler and thus better first example:
Screen Shot 2021-09-19 at 12 23 34 PM

leem44:
@ttimbers Changes look good! Only thing is the one below "Montreal" is spelled "Montreak" 

> Also changed the vector figure to be of type character, as previously it was called an integer, but then we actually made a double vector in code... I tried to swap it to double, but then it was too confusing for a first example. Character vector is a simpler and thus better first example:
> Screen Shot 2021-09-19 at 12 23 34 PM

ttimbers:
Good catch!

ttimbers:
I also removed loading data from the canlang package and instead get them to read the data from files. I don't think we really explain data packages and they are kind of weird, so best to leave out I think?

leem44:
Yes, I agree with that! Makes sense!

ttimbers:
In Fig 3.10 we illlustrate going from longer to wider, but each table is the same number of columns. Yes, this is technically something that happens, but for our first teaching case we really should have a wider dataframe that we make longer:
Screen Shot 2021-09-21 at 10 26 13 PM

Also, where did the values for commuters come from? I imagine population came from `canlang`, but I don't think that data set has commuters? Other columns we could add from `canlang` could be area? I will try to do that.

ttimbers:
I think we should remove the `&` operator from here as we never use it elsewhere in the book or course. So it's just redundant and extra info that they do not "need" to know:
Screen Shot 2021-09-22 at 9 48 06 AM

ttimbers:
@leem44 - I am also removing the demonstration of `filter` not working on numeric data that is in the `mutate` section, as it is rendundant to what we wrote above in the section about `convert` and `separate`.

ttimbers:
@leem44 - in the wrangling chapter, are you OK to move the pipe to before `mutate` and after `select` and `filter`? It uses `select` and `filter` so is timely after those two. Also, then I can use the `|>` in the setup for the `as.numeric` transformation for `mutate` so the creation of the data frame with the character columns that should be numbers is a little less magical?

ttimbers:
Simplified the example to just plot the proportion of people speaking English as the primary language at home.
Then we can skip case_when and vector recycling and keep the narrative pretty similar.

ttimbers:
Done all sections except for `summarize` + `across` and `purrr::map*` (I did do `rowwise` which comes after these).

ttimbers:
OK, I think we are ready to merge.  I added a bunch of images into the wrangling chapter, and re-organized the section where we do aggregation. I am pretty excited about it. @leem44 - I made quite a few changes to the wrangling chapter to (hopefully) simplify it. I hope that is OK. If there is anything I removed that you really want to keep in, please let me know and we can talk and find a way to keep it.

ttimbers:
@leem44 gave me the go-ahead to merge, so merging!

Issue #96, opened 2021-03-22
Review: Global/big picture formatting revisions


ttimbers:

leem44:
From reviewer E: 
- placement of figures is confusing throughout this text (e.g. Sections 1.62, 4.5.2). figures are often far from what they described; sometimes even directly under the wrong code chunk. 
  - recommend forcing appropriate figure placement or, at minimum, using more specific language of “In Figure X, we see…” instead of “in the plot below”.
- code frequently spills out over the textboxes. may need to split some commands across more lines or change the font.

trevorcampbell:
# Reviewer D
- background colour used for the figures is not always consistent.  Some figures use the default grey theme of ggplot, while others don’t.  The grey theme is not the best default so perhaps the authors could consider using a more visually appealing theme.  
- There aren’t many tables in the manuscript and I think this is a missed opportunity. I could only find one table, which lists the questions typical of various kinds of statistical analyses but even this was borrowed from a different source and included in the manuscript without much of a preamble.  Tables are a great way to show information side-by-side (e.g., pros and cons of a statistical procedure) or to list various options the students can consider (e.g., different types of geoms that could be used to build a ggplot; different types of R import functions that could be used to read in different types of data files). It would be good if the authors could add a few more tables to the manuscript. 
    - This is not a bad idea, especially for chapters with a lot of such "options" (viz is a great example)
    - Just to be clear: tables should not be used for conveying numerical/statistical results/comparisons. Figures are always better.

trevorcampbell:
# Reviewer A
- the color schemes should be carefully chosen so that colorblind readers have no problem reading them. Double encoding could also be used to help with this.
- From a typesetting perspective, there are some problems with the figures in the PDF (but not the HTML version). In the PDF, the figures often fall too far from the discussion or code snippet. There is certainly some tweaking to be done in typesetting, but I also think that it is important to explicitly point to the figure being discussed for the print edition. 
- the book is of the right length for an introduction
- be sure no code gets cut off and that figures are placed as close to the text as possible.

trevorcampbell:
# Trevor
- we say "Professor Claus Wilke" and "Grolemund & Wickham" when referring to textbooks. We should go through and make sure we reference other resources in a consistent way

trevorcampbell:
### Edit this comment directly to summarize / synthesize the major (important) comments across reviewers. If there's already the same comment below, add the reviewer to the parenthetical list at the beginning.

[...]

PR #237, opened 2021-08-19
Review installation chapter

leem44:
@ttimbers @trevorcampbell 
review of the installation chapter! 
There were a lot of links in this chapter, which how they are currently displayed won't make sense in the print book, so we should think about how we want those to display in the book (Maybe just type out the URL beside?). I added it to #96 as something we need to fix!


ttimbers:
I just reviewed this and I:
- updated and simplified installation instructions (and tested them on my mac)
- moved the moving files from JupyterHub to an appendix

I think we can merge this now?

PR #238, opened 2021-09-22
Create an index, style guide, handle remaining bugs

### LINKED ISSUE #120: Fix/improve index
We need a robust index for this book. Check whether this can be autogenerated by `bookdown`? Talk to Laura (CRC Press) for help on this if we need to? Also, once we create the index, we want to create a glossary of the main terms and functions. Let's consider using the [`glossario` R package](https://github.com/carpentries/glosario) for this, and borrowing from the [Carpentries English glossary](https://glosario.carpentries.org/en/)? This addresses comments made by Rev E.  **MINOR/MAJOR?**

### LINKED ISSUE #168: set.seed once at beginning of each chapter when loading packages
currently we set seeds throughout chapters in kind of a haphazard way. Probably should just set the seed once at the beginning of each chapter. We also have a few "note" boxes throughout about seeds, should edit those once seed moved to beginning of chp.

### LINKED ISSUE #187: data base for ch 2
Mentioned by Reviewer A in chapter 2: 
- p 39: The reader won't be able to connect to this DB, so warn them!
     - TC: we should just point them to a DB they *can* connect to...

_Originally posted by @trevorcampbell in https://github.com/UBC-DSCI/introduction-to-datascience/issues/102#issuecomment-808957676_

### LINKED ISSUE #189: videos in chapter 2
Just flagging a comment from Reviewer A about chapter 2 video so we don't miss it: 
- page 33: The video says you can click on the data to open it in the editor, but this doesn't match the text. Be sure to update the video before publication

_Originally posted by @trevorcampbell in https://github.com/UBC-DSCI/introduction-to-datascience/issues/102#issuecomment-808957676_

### LINKED ISSUE #195: Feedback on "Multicollinearity and outliers" section in linear regression section
- [ ] The first paragraph uses terms we have not used in the book before (outliers and collinear predictors) without defining them. Could we roughly hint at what they are here, or say that we will explain them, and their impact below?
- [ ] At the end of the outlier section, can we point to resources where the can read more about this in case they are interested or have to deal with this?
- [ ] In the outlier section, should we discuss the simple solution of dropping outliers? Why this can work sometimes, and why sometimes this should be avoided and instead other advanced methods (like median or quantile regression, should be used?)
- [ ] What do unstable coefficients mean, we should explain this term.
- [ ] We say "Therefore, when performing multivariate linear regression, it is important to avoid including very linearly related predictors." but this begs the question how do we avoid this? Could we suggest or show some simple ways of looking at this? And then what to do if they see this? For example, in the case of the two measurements, you could drop one, or you could create a new feature from these two where you average them?
- [ ] At the end of the multicollinearity section, can we point to resources where the can read more about this in case they are interested or have to deal with this in a more complex case than we present?

### LINKED ISSUE #196: Feedback on "Designing new predictors"
- [ ] We write "For example, a dataframe df with two variables—x and y—with a nonlinear relationship between the two variables will not be correctly captured by simple linear regression, as shown in Figure 9.11. " but is "correct" the right word? Maybe "fully" might be better?
- [ ] Can we make it clear that we drop X, when we add Z to the model?
- [ ] Should we use captitals to refer to X, Y and Z here?
- [ ] what is `df`, we never define it... 
- [ ] At the end of this section, can we point to resources where the can read more about this in case they are interested or have to deal with this in a more complex case than we present?

### LINKED ISSUE #236: Clustering chapter figure
The figure below 9.7 is supposed to be a bad clustering but it looks fine

### LINKED ISSUE #240: Formatting of $K$
I agree with this change in general. But I have noticed in the HTML navigation table that appears on the left, the $K$ symbol is really, really small.

My guess is that the PDF output looks fine, but we should check it.

I am not sure this is an issue we should fix, since it seems to be a problem with how rbookdown renders math in the navigation pane.

_Originally posted by @trevorcampbell in https://github.com/UBC-DSCI/introduction-to-datascience/pull/234#discussion_r713572639_

trevorcampbell:
This PR is not done yet. 

I've finished everything listed below except the index (waiting for wrangling PR).

To address:
- #120 
- #168   
- #187   
- #195   
- #196     
- #240  
- #236  
- #189 

### Index

I made one! This required some fixes to enable compile to PDF:

**Fixes to make the book compile to PDF**
- need to add figure units to `out.width`
    - done: added `pt`
- the gentoo penguin image in clustering chapter breaks things with percent signs in its name
    - done: i stored a local copy of it instead
- clustering two-column "center update / label update" spaces use unicode (non-latex) characters
    - done: no-op. `xelatex` handles unicode characters.
- `img/population_vs_sample.svg`  -- svgs break pdf output
    - done: i rendered this to png instead
- unicode characters in the jupyter section
     - done: no-op. `xelatex` handles unicode characters.
- duplicate entry in refs.bib (`wilson2014best`)
     - done: removed old entry
- Lots of `fig.retina = ...` in `inference.Rmd` without actually setting `out.width` (causes LaTeX to generate `width=468px` which breaks `xelatex` because the unit should be `pt`
    - done: either set width explicitly or removed setting retina
- Fixed `@online` entries in `references.bib` to be `@misc`

This work closes #120 

### Style Guide
I created one. It's in README.md. I just made executive decisions on everything in the interest of moving this along, but feel free to raise anything I did there as a discussion item. Also feel free to move the style guide into its own `.md` file and link from `README.md` if desired.

The work on the style guide closes #240 -- but note that we still need to do the actual formatting pass to fix it to conform to the style.

### Clustering
Fixed the two examples of "failures" in clustering (bad init, bump in elbow plot).
This closes #236 

### Videos in chapter 2
No-op. These videos were removed (I assume it was the jupyter one). The only videos left are in the additional resources at the end, and those aren't about previewing data in jupyter. The jupyter chapter no longer has any videos.

I've added to the style guide a mention of making sure video links are usable in both pdf and html.
This work closes #189

### Database in Chapter 2
Made it clearer that the DB host/username/password info that we use in ch2 is fake. This closes #187 

agree, done

 > In the outlier section, should we discuss the simple solution of dropping outliers? Why this can work sometimes, and why sometimes this should be avoided and instead other advanced methods (like median or quantile regression, should be used?)

Nope - I don't think we should go into it in this text. I don't think it's possible to convey it in such a way that it won't cause students to go "oh well I don't like this datapoint, it looks *wrong* to me" and then just arbitrarily throw out data. 

 > What do unstable coefficients mean, we should explain this term.

I agree, unstable is a bit too technical -- replaced with "unreliable" + explanation instead!

> We say "Therefore, when performing multivariate linear regression, it is important to avoid including very linearly related predictors." but this begs the question how do we avoid this? Could we suggest or show some simple ways of looking at this? And then what to do if they see this? For example, in the case of the two measurements, you could drop one, or you could create a new feature from these two where you average them?

Definitely out of scope. I just point the reader to the end of the chapter.

> At the end of the multicollinearity section, can we point to resources where the can read more about this in case they are interested or have to deal with this in a more complex case than we present?

See above.

> At the end of the outlier section, can we point to resources where the can read more about this in case they are interested or have to deal with this?

agreed, but not worth the effort before CRC subm

This work closes #195 

### Feedback on predictor design

>  We write "For example, a dataframe df with two variables—x and y—with a nonlinear relationship between the two variables will not be correctly captured by simple linear regression, as shown in Figure 9.11. " but is "correct" the right word? Maybe "fully" might be better?

agree, done

>  Can we make it clear that we drop X, when we add Z to the model?

I don't think we need to -- the plot axes are clear that we're plotting z now

>  Should we use captitals to refer to X, Y and Z here?

I don't see any reason to - lowercase seems fine

>  what is df, we never define it...

it was defined in the text, but I added a quick printout of df at the beginning

>  At the end of this section, can we point to resources where the can read more about this in case they are interested or have to deal with this in a more complex case than we present?

I agree, but not worth the effort prior to CRC subm.

This work closes #196 

### `set.seed`
- seed/randomness was previously first introduced at the end of classification 1 for a single off-the-cuff usage in picking 3 observations to create imbalanced data; I thought it was a bit hidden and not clearly described
- I removed that usage of randomess
- now the first introduction of randomness is in classification2, where we can justify it clearly saying "OK now we need randomness since we're splitting data"
- I have now devoted a whole section to randomness and seeds
- it's now clear that you only set.seed once at the beginning of your analysis
- I've hidden everywhere in the textbook where we "seed hack" to get results we want (for pedagogical reasons) and just put a seed at the start of the chapter when we load tidyverse, reminding the reader why we do this

This work closes #168 

### Miscellaneous
- fixed y axis labels in inference chapter
- fixed typos throughout
- cleaned the appendix a bit
- fixed weird learning objective spacing in chapter 2
- fixed super long lines in ch2; 80ch line limit
- moved `faithful_plot.*` images to `img/` folder
- fixed bad sample in the first example (`small_sacramento`) in regression1 (there was a point lying basically right on top of the dashed line)


inference.Rmd
```diff
@@ -224,12 +235,13 @@ entire home/apartment listings in each sample. This simulation will create
 many sample proportions, which we can visualize using a histogram. The
```
leem44, line 235: above add "sampling variability" to index

inference.Rmd
```diff
@@ -224,12 +235,13 @@ entire home/apartment listings in each sample. This simulation will create
...
 expect our sample proportions from this population to vary for samples of size 40. 
 
 We again use the `rep_sample_n` to take samples of size 40 from our
 population of Airbnb listings. But this time we set the `reps` argument to 20,000 to specify 
-that we want to take 20,000 samples of size 40.
+that we want to take 20,000 samples of size 40. \index{rep\_sample\_n!reps argument}
+\index{rep\_sample\_n!size argument}
 ```{r 11-example-proportions5, echo = TRUE, message = FALSE, warning = FALSE}
 samples <- rep_sample_n(airbnb, size = 40, reps = 20000)
 samples
```
leem44, line 247: below add `tail` to index

inference.Rmd
```diff
@@ -328,7 +342,7 @@ population_parameters
 ```
 The price per night of all Airbnb rentals in Vancouver, BC
```
leem44, line 343: above this: 
Screen Shot 2021-09-26 at 9 08 42 PM

inference.Rmd
```diff
@@ -328,7 +342,7 @@ population_parameters
 ```
 The price per night of all Airbnb rentals in Vancouver, BC 
 is \$`r round(population_parameters$pop_mean,2)`, on average. This value is our
-**population parameter** since we are calculating it using the population data. 
+**population parameter** since we are calculating it using the population data. \index{population!parameter}
```
leem44, line 345: We should probably remove the bolding here since we define it in the first paragraph of this chapter already

inference.Rmd
```diff
@@ -373,7 +389,7 @@ took another random sample from the population, our estimate's value might
 change. So then, did we just get lucky with our point estimate above?  How much
 does our estimate vary across different samples of size 40 in this example?
 Again, since we have access to the population, we can take many samples and
-plot the **sampling distribution** of sample means for samples of size 40 to
+plot the **sampling distribution** of \index{sampling distribution} sample means for samples of size 40 to
```
leem44, line 392: here remove bolding since "sampling distribution" was defined earlier

inference.Rmd
```diff
@@ -663,7 +691,7 @@ sample *is* the population, and take more samples (with replacement) of the
 same size from it instead! This very clever technique is 
 called **the bootstrap**.  Note that by taking many samples from our single, observed
 sample, we do not obtain the true sampling distribution, but rather an
-approximation that we call **the bootstrap distribution**.
```
leem44, line 666: above remove bolding for the word "point estimate" at very start of this section since it was defined already

inference.Rmd
```diff
@@ -702,6 +730,7 @@ one_sample <- one_sample |>
 one_sample
 one_sample_dist <- ggplot(one_sample, aes(price)) +
```
leem44, line 731: need to fix the figure formatting here 
Screen Shot 2021-09-26 at 9 15 58 PM

version-control.Rmd
```diff
@@ -125,28 +125,28 @@ and repository hosting services in use today.
...
-This copy is commonly referred to as the **remote repository**.
+This copy is commonly referred to as \index{repository!remote} the **remote repository**.
 
 ```{r vc1-no-changes, fig.cap = 'Schematic of local and remote version control repositories.', fig.retina = 2}
 image_read("img/vc1-no-changes.png") %>%
   image_crop("3632x2000")
 ```
 
-Both copies of the repository have a **working directory**
+Both copies of the repository have a **working directory**
```
leem44, line 142: add "working directory" to index?

leem44 (commented):
@trevorcampbell Thank you for putting all of this together!! Looks great! 
- The page numbers in the index seem to be off by 2. E.g. "data science": in index says it's on page 5 but really its on page 7 
- I guess we'll need to do another pass and look at all the plots and code that fall off page, and also the plots that are in weird places (e.g. in the middle of text etc)
- nice addition of the `set.seed` explanation!
- a bunch of the figure references display "Figure ??" -- I started pointing them out and then realized that they are in a lot of places so likely this will need to be a part of the formatting pass

intro.Rmd
```diff
@@ -144,26 +149,26 @@ according to the 2016 census.
...
 Since we are using R for data analysis in this book, the first step for us is to
 load the data into R. When we load tabular data into
-R, it is represented as a *data frame* object. Figure
+R, it is represented as a *data frame* object\index{data frame!overview}. Figure
 \@ref(fig:img-spreadsheet-vs-dataframe) shows that an R data frame is very similar
-to a spreadsheet. We refer to the rows as **observations**; these are the things that we
+to a spreadsheet. We refer to the rows as **observations** \index{observation}; these are the things that we
 collect the data on, e.g., voters, cities, etc. We refer to the columns as
-**variables**; these are the characteristics of those observations, e.g., voters' political
+**variables** \index{variable}; these are the characteristics of those observations, e.g., voters' political
```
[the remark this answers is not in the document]
trevorcampbell: fixed

intro.Rmd
```diff
@@ -144,26 +149,26 @@ according to the 2016 census.
...
 rectangular-shaped and spreadsheet-like, as shown in Figure
 \@ref(fig:img-spreadsheet-vs-dataframe). In this book, we will focus primarily on tabular data.
 
 Since we are using R for data analysis in this book, the first step for us is to
 load the data into R. When we load tabular data into
-R, it is represented as a *data frame* object. Figure
+R, it is represented as a *data frame* object\index{data frame!overview}. Figure
 \@ref(fig:img-spreadsheet-vs-dataframe) shows that an R data frame is very similar
-to a spreadsheet. We refer to the rows as **observations**; these are the things that we
+to a spreadsheet. We refer to the rows as **observations** \index{observation}; these are the things that we
```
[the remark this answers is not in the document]
trevorcampbell: fixed

intro.Rmd
```diff
@@ -476,7 +482,7 @@ ggplot(ten_lang, aes(x = language, y = mother_tongue)) +
 > end with a `+` symbol, and so it keeps reading the next line to figure out
```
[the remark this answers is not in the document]
trevorcampbell: Yes! fixed

inference.Rmd
trevorcampbell: fixed

inference.Rmd
trevorcampbell: fixed

inference.Rmd
trevorcampbell: done

version-control.Rmd
trevorcampbell: yes!

intro.Rmd
```diff
@@ -76,7 +76,13 @@ descriptive, exploratory, inferential, predictive, causal, and mechanistic,
 all of which are defined in Table \@ref(tab:questions-table).
 Carefully formulating a question as early as possible in your analysis&mdash;and 
 correctly identifying which type of question it is&mdash;will guide your overall approach to 
-the analysis as well as the selection of appropriate tools.
+the analysis as well as the selection of appropriate tools.\index{question!data analysis}
+\index{descriptive!definition}
```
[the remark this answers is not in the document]
trevorcampbell: done

intro.Rmd
```diff
@@ -251,14 +256,15 @@ What would be more useful would be to give a name
 to the data frame that `read_csv` outputs, 
 so that we can refer to it later for analysis and visualization.
 
-The way to assign a name to a value in R is via the *assignment symbol* `<-`.
+The way to assign a name to a value in R is via the *assignment symbol* `<-`. 
+\index{leftarrow@\texttt{<-}|see{assignment symbol}}\index{assignment symbol}
```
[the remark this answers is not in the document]
trevorcampbell: I think you may have misread the entry -- the `<-` symbol is on its own, it isn't a subentry.

But I agree that the ordering of symbols in the index is a bit unclear. I'll make an issue to think about handling it

inference.Rmd
trevorcampbell: created new issue for it

PR #250, opened 2021-09-27
Clustering

ttimbers:
@trevorcampbell @leem44 - this PR is the final review of the clustering chapter. It should close #110 . Can one of you please do a review of this? At least a light one?

PR #252, opened 2021-09-28
Addressed @trevorcampbell 's unaddressed comments from #250

ttimbers:


clustering.Rmd
```diff
@@ -151,30 +154,31 @@ penguin_data <- read_csv("data/penguins_standardized.csv")
...
+we might suspect there are a few subtypes of penguins within our data set.
+We can see roughly 3 groups of observations in Figure \@ref(fig:10-toy-example-plot),
 including:
 
 1. a small flipper and bill length group,
 2. a small flipper length, but large bill length group, and
 3. a large  flipper and bill length group.
 
-This is interesting that we can use data visualization to uncover such patterns,
+It is interesting that we can use data visualization to uncover such patterns,
```
trevorcampbell, line 178: This sentence isn't grammatically correct

Let's go with

```
Data visualization is a great tool to give us a rough sense for such patterns
when we have a small number of variables. But if we are to group
data&mdash;and select the number of groups&mdash;as part of 
a reproducible analysis, we need something a bit more automated.
````

File: clustering.Rmd [hunk 2]
```diff
@@ -663,11 +671,11 @@ penguin_clust_ks
...
 
 Then we use `rowwise` \index{rowwise} + `mutate` to apply the `kmeans` function 
-within each row, to each K. 
+within each row to each K. 
 However, given that the `kmeans` function 
-returns a model object to us, (not a vector),
+returns a model object to us (not a vector)
 we will need to store the results as a list columm.
-This can work because lists are super vectors, and so they are legitimate 
+This can work because both vector and lists are legitimate
```
trevorcampbell, line 678: "model object to us (not a vector)" -> "model object to us (not a vector),"    (added comma at the end)

"This can work" -> "This works"

"both vector" -> "both vectors"

clustering.Rmd
ttimbers: And then I am going to add this point after that as well: 

```
Additionally, finding groups through data visualization becomes more difficult 
as we increase the number of variables we consider when clustering.
```

clustering.Rmd
ttimbers: fixed

PR #263, opened 2021-10-01
Viz

### LINKED ISSUE #104: Review: Ch 4 (viz)
# Reviewer E: 
- [x] Section 4.5.4 the formula interface is used with facet_grid(). I believe the preferred approach now is to use the ‘rows’ and ‘cols’ parameters instead. 
- [x] Ch title/subtitles: The actual titles are fine, but for some reason there are unneeded extra levels in the subsection numbers. This is immaterial but unaesthetic

- ~~I really like the introduction. The only thing I thought could be expanded on was the distinction between exploratory and explanatory visualization and how one might approach visualization as part of EDA versus visualizations to document a result~~
- [x] I struggled with the data in the example here. What are the date versus date_decimal fields? 
- [x] When reading in the Manua Loa data, no explanation is given for the filter statement (ppm>0). This was somewhat distracting, so it might be worth either explaining or preprocessing the data to have the correct subset before it is read in. 
- [x] The theme() function is very complex and has many potential arguments. I would recommend either taking the time to explain how it works in more detail or removing it from the text. I worry showing one single example without explanation may lead students to a misunderstanding of what it is.
- ~~In section 4.5.2, the question asked of the plot could easily be answered by a table. Is there a risk here of making students feel that they must plot everything? More broadly, should how to make a good table be covered in parallel as another aspect of data visualization?~~
- [x] Consider showing how to modify an existing ggplot2 object instead of repeatedly recreating (e.g. islands_bar + coord_flip() instead of repeating code to make islands_bar)
- [x] Love the section on explaining a visualization! 
- ~~The section on saving plots seems long and unusually detailed (e.g. raster versus vector images), and I might consider removing. This is all interesting stuff, but it feels very in-depth compared to the higher level treatment of core data analysis and modeling principles. In particular, the book assumes students are working in notebooks where saving images is generally not needed.~~

ttimbers:
Addresses reviewers comments in #104 except where I stated I would not address them.

Closes #104


viz.Rmd
```diff
@@ -85,82 +94,172 @@ making it easier for them to understand and remember your message quickly.
...
+> (linked to in our Additional Resources section at the end of this chapter).
+
+Since we are investigating a relationship between two variables 
+(CO$_{\text{2}}$ concentration and date), 
+a scatter plot is a good place to start. 
+Scatter plots show the data as individual points with `x` (horizontal axis) 
+and `y` (vertical axis) coordinates.
+An example of a scatter plot is shown in Figure \@ref(fig:scatter-plot-sketch).
+
+```{r scatter-plot-sketch, echo = FALSE, fig.width = 3.25, fig.height = 3, fig.align = 'center', fig.cap = "Example of a scatter plot."}
```
leem44, line 188: I really like these plots you created and I think these are so great for explaining the basic plots types! I think we should move all of these plots up to section 4.3: Choosing the visualization. Right after we introduce the list of plots we can give a visual representation of the plots with these figures you've created. It makes the list in 4.3 less abstract and also allows people to compare the visualizations. It also makes these sections less distracting so we can focus only on the question  we want to answer about the data + iterating as opposed to learning what a scatterplot is *and* trying to answer the question

viz.Rmd
```diff
@@ -55,7 +64,7 @@ As with most coding tasks, it is totally fine (and quite common) to make mistake
 
 - **line plots** visualize trends with respect to an independent, ordered quantity (e.g. time)
 - **histograms** visualize the distribution of one quantitative variable (i.e., all its possible values and how often they occur) 
-- **scatter plots** visualize the distribution / relationship of two quantitative variables
+- **scatter plots** visualize the relationship between two quantitative variables
 - **bar plots** visualize comparisons of amounts
 
 All types of visualization have their (mis)uses, but three kinds are usually hard to understand or are easily replaced with an oft-better alternative.
```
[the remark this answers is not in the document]
leem44: Edit: I think if we move the plots you created up to this section (see my comment below) then I don't think we need to move this. Since we can show the pictures then put this paragraph underneath the pictures

viz.Rmd
```diff
@@ -85,82 +94,172 @@ making it easier for them to understand and remember your message quickly.
...
+> **Note:** `read_csv` was able to parse the `date_measured` column into the
+> `date` vector type because it was entered 
+> in the international standard date format, 
+> called ISO 8601, which lists dates as `year-month-day`.
+> `date` vectors are `double` vectors with special properties that allow 
+> them to handle dates correctly.
+> For example, `date` type vectors allow functions like `ggplot` 
+> to treat them as numeric dates and not a character vectors, 
+> even though they contain non-numeric characters 
+> (e.g., `-` in the `date_measured` column in the `co2_df` data frame).
```
leem44, line 173:
```suggestion
> (e.g., the dash (`-`) in the `date_measured` column in the `co2_df` data frame).
```
in the pdf its hard to see the "-" character

viz.Rmd
```diff
@@ -85,82 +94,172 @@ making it easier for them to understand and remember your message quickly.
...
+The `date_measured` column holds the date the measurement was taken, 
+and is of type `date`.
+The `ppm` columns holds the value of $CO_2$ in parts per million 
+that was measured on each date, and is type `double`.
+
+> **Note:** `read_csv` was able to parse the `date_measured` column into the
+> `date` vector type because it was entered 
+> in the international standard date format, 
+> called ISO 8601, which lists dates as `year-month-day`.
+> `date` vectors are `double` vectors with special properties that allow
```
leem44, line 168: I think we should move the line : "date vectors are double vectors with special properties that allow them to handle dates correctly. For exam- ple, date type vectors allow functions like ggplot to treat them as numeric dates and not a character vectors, even though they contain non-numeric characters (e.g., - in the date_measured column in the co2_df data frame). " to the beginning of this note and then say "read_csv was able to parse..."

viz.Rmd
```diff
@@ -85,82 +94,172 @@ making it easier for them to understand and remember your message quickly.
...
+that was measured on each date, and is type `double`.
+
+> **Note:** `read_csv` was able to parse the `date_measured` column into the
+> `date` vector type because it was entered 
+> in the international standard date format, 
+> called ISO 8601, which lists dates as `year-month-day`.
+> `date` vectors are `double` vectors with special properties that allow 
+> them to handle dates correctly.
+> For example, `date` type vectors allow functions like `ggplot` 
+> to treat them as numeric dates and not a character vectors,
```
leem44, line 171:
```suggestion
> to treat them as numeric dates and not character vectors, 
```

viz.Rmd
```diff
@@ -85,82 +94,172 @@ making it easier for them to understand and remember your message quickly.
...
+> `date` vector type because it was entered 
+> in the international standard date format, 
+> called ISO 8601, which lists dates as `year-month-day`.
+> `date` vectors are `double` vectors with special properties that allow 
+> them to handle dates correctly.
+> For example, `date` type vectors allow functions like `ggplot` 
+> to treat them as numeric dates and not a character vectors, 
+> even though they contain non-numeric characters 
+> (e.g., `-` in the `date_measured` column in the `co2_df` data frame).
+> This means R it will not accidentally plot the dates in the wrong order
```
leem44, line 174:
```suggestion
> This means R will not accidentally plot the dates in the wrong order 
```

viz.Rmd
```diff
@@ -85,82 +94,172 @@ making it easier for them to understand and remember your message quickly.
...
+> called ISO 8601, which lists dates as `year-month-day`.
+> `date` vectors are `double` vectors with special properties that allow 
+> them to handle dates correctly.
+> For example, `date` type vectors allow functions like `ggplot` 
+> to treat them as numeric dates and not a character vectors, 
+> even though they contain non-numeric characters 
+> (e.g., `-` in the `date_measured` column in the `co2_df` data frame).
+> This means R it will not accidentally plot the dates in the wrong order 
+> (i.e., not alphanumerically as would happen if it was a character vector). 
+> An in depth study of dates and times is beyond the scope of the book,
```
leem44, line 176:
```suggestion
> An in-depth study of dates and times is beyond the scope of the book, 
```

viz.Rmd
```diff
@@ -85,82 +94,172 @@ making it easier for them to understand and remember your message quickly.
...
+> For example, `date` type vectors allow functions like `ggplot` 
+> to treat them as numeric dates and not a character vectors, 
+> even though they contain non-numeric characters 
+> (e.g., `-` in the `date_measured` column in the `co2_df` data frame).
+> This means R it will not accidentally plot the dates in the wrong order 
+> (i.e., not alphanumerically as would happen if it was a character vector). 
+> An in depth study of dates and times is beyond the scope of the book, 
+> but we invite readers who are interested to learn more 
+> to see the Dates and times chapter of R for Data Science [@wickham2016r]
+> (linked to in our Additional Resources section at the end of this chapter).
```
leem44, line 179:
```suggestion
> (linked in our additional resources section at the end of this chapter).
```
are we doing lower case or upper case "Additional Resources"? I think Trevor corrected mine to lower case on the last edit

viz.Rmd
```diff
@@ -55,7 +64,7 @@ As with most coding tasks, it is totally fine (and quite common) to make mistake
 
 - **line plots** visualize trends with respect to an independent, ordered quantity (e.g. time)
```
leem44, line 65: I think we should make this list align with the order that we introduce them in the chapter
Also should we group the graphs? E.g. put all the scatterplots together etc (co2 data + eruption + canlang)

viz.Rmd
```diff
@@ -85,82 +94,172 @@ making it easier for them to understand and remember your message quickly.
...
 
-Since we are investigating a relationship between two variables (CO2 concentration and date), a scatter plot is a good place to start. Scatter plots 
-show the data as individual points with `x` (horizontal axis) and `y` (vertical axis) coordinates. Here, we will use the decimal
- date as the `x` coordinate 
-and CO2 concentration as the `y` coordinate. When using the `ggplot2` package, we create the plot object with the `ggplot` function; there are 
-a few basic aspects of a plot that we need to specify:
+We see we have two columns available to us; `date_measured` and `ppm`. 
+The `date_measured` column holds the date the measurement was taken, 
+and is of type `date`.
+The `ppm` columns holds the value of $CO_2$ in parts per million
```
leem44, line 161:
```suggestion
The `ppm` column holds the value of CO$_{\text{2}}$ in parts per million 
```

viz.Rmd
```diff
@@ -85,82 +94,172 @@ making it easier for them to understand and remember your message quickly.
...
+
+ggplot(line_data, aes(x = x, y = y)) +
+  geom_line() +
+  xlim(c(0, 12)) +
+  ylim(c(0, 12)) +
+  theme_classic()
+```
+
+
+Let's now try tp visualize the `co2_df` as a line plot
```
leem44, line 257:
```suggestion
Let's now try to visualize the `co2_df` as a line plot 
```

viz.Rmd
```diff
@@ -176,47 +275,123 @@ To add axis labels, we use the `xlab` and `ylab` functions. To change the font s
...
+co2_line <- ggplot(co2_df, aes(x = date_measured, y = ppm)) +
   geom_line() +
   xlab("Year") +
   ylab("Atmospheric CO2 (ppm)") +
-  theme(text = element_text(size = 18))
+  theme(text = element_text(size = 16))
 co2_line
 ```
 
+> **Note:**
```
leem44, line 287: In the pdf this note is misplaced, it appears under the figure so its a little confusing -- this is probably something we need to do in a larger formatting pass (placement of figures and text), but noting it here just in case
Screen Shot 2021-09-30 at 9 30 03 PM

viz.Rmd
```diff
@@ -176,47 +275,123 @@ To add axis labels, we use the `xlab` and `ylab` functions. To change the font s
...
 co2_line
 ```
 
+> **Note:** 
+> The `theme` function is quite complex and has many arguments 
+> that can be specified to control many non-data aspects of a visualization.
+> An in-depth discussion of the `theme` function is beyond the scope of this book,
+> however we invite readers who are interested to learn more 
+> to read the `theme` function documentation
+> (linked to in our Additional Resources section at the end of this chapter).
```
leem44, line 293:
```suggestion
> (linked to in our additional resources section at the end of this chapter).
```

leem44 (commented):
@ttimbers Really nice improvements to this chapter!!! The new plots are excellent! A few comments below

viz.Rmd
trevorcampbell: "additional resources section"

"Section X"

are current style

trevorcampbell:
@ttimbers just skimming this -- need to update the learning objectives (e.g. `scale_x/y_continuous` doesn't appear any more, and I think some new funcs are there now!)

viz.Rmd
```diff
@@ -529,11 +770,11 @@ morley_hist <- ggplot(morley, aes(x = Speed, fill = Expt)) +
 morley_hist
```
leem44, line 770: do we explain `position = "identity"` -- I think we started using it here, but don't think it's explained... I think we should remove it here since it doesn't do anything in this particular line. 
But we should explain it below when the colours are introduced below

viz.Rmd
```diff
@@ -529,11 +770,11 @@ morley_hist <- ggplot(morley, aes(x = Speed, fill = Expt)) +
...
 ```
 
-Wait a second; the histogram is still all the same colour! What is going on here? If we look at the printed `morley` data, the column `Expt` is an integer (we see the label `<int>` underneath the `Expt` column name). But we want to treat it as a categorical variable.  To fix this issue we can write `factor(Expt)` in the `fill` aesthetic mapping. Factors are a special categorical type of variable in R that are often used for class label data. By writing `factor(Expt)` we are ensuring that R will treat this variable as a factor and the colour will be mapped discretely.
+Wait a second; the histogram is still all the same colour! What is going on here? If we look at the printed `morley` data, the column `Expt` is an integer (we see the label `<int>` underneath the `Expt` column name). But we want to treat it as a categorical variable.  To fix this issue we can write `as_factor(Expt)` in the `fill` aesthetic mapping. Recall that factors are a special categorical type of variable in R that are often used for class label data. By writing `as_factor(Expt)` we are ensuring that R will treat this variable as a factor and the colour will be mapped discretely.
```
leem44, line 773:
```suggestion
Wait a second; the histogram is still all the same colour! What is going on here? If we look at the printed `morley` data, the column `Expt` is an integer (we see the label `<int>` underneath the `Expt` column name). But we want to treat it as a categorical variable.  To fix this issue, we can write `as_factor(Expt)` in the `fill` aesthetic mapping. Recall that factors are a special categorical type of variable in R that are often used for class label data. By writing `as_factor(Expt)` we are ensuring that R will treat this variable as a factor, and the colour will be mapped discretely.
```

viz.Rmd
```diff
@@ -547,16 +788,26 @@ are blending together, and although it is possible to derive *some* insight from
...
+The argument to `facet_grid` specifies the variable(s) used to split the plot 
+into subplots, and how to split them (i.e., into rows or columns).
+If the plot is to be split horizontally, into rows, 
+then the `rows` argument is used.
+If the plot is to be split vertically, into columns, 
+then the `columns` argument is used.
+Both the `rows` and `columns` argument take the column names to split the data 
+on when creating the subplots. 
+One key thing is that the column names must be surrounded by the `vars` function.
+This function allows the column names to be correctly evaluated
```
leem44, line 803: could we explain the `vars` function a bit more here

viz.Rmd
```diff
@@ -547,16 +788,26 @@ are blending together, and although it is possible to derive *some* insight from
...
-```{r 03-data-morley-hist-4, warning=FALSE, message=FALSE, fig.cap = "Histogram of Michelson's speed of light data split vertically by experiment"}
-morley_hist <- ggplot(morley, aes(x = Speed, fill = factor(Expt))) +
+```{r 03-data-morley-hist-4, warning=FALSE, message=FALSE, fig.height = 7, fig.cap = "Histogram of Michelson's speed of light data split vertically by experiment"}
+morley_hist <- ggplot(morley, aes(x = Speed, fill = as_factor(Expt))) +
   geom_histogram(position = "identity") +
-  facet_grid(Expt ~ .) +
+  facet_grid(rows = vars(Expt)) +
   geom_vline(xintercept = 792.458, linetype = "dashed", size = 1.0)
 morley_hist
 ```
```
leem44, line 813: I think after the line: "makes it quite clear how accurate the different experiments were with respect to one another. "
I think it would help the reader to provide a quick interpretation of the plot e.g. "We can experiment 1 had the highest spread/variability (whatever term we have used in the past) and experiment 6 blah etc."

viz.Rmd
```diff
@@ -569,23 +820,115 @@ for just how accurate all the experiments were overall. For example, how accurat
 To answer this question, we'll use the `mutate` function to transform our data into a relative measure of accuracy rather than absolute measurements:
```
leem44, line 820: in the line above: "it is hard to get a sense for just" -- change "for" to "of"

viz.Rmd
```diff
@@ -569,23 +820,115 @@ for just how accurate all the experiments were overall. For example, how accurat
...
+as well as the number of observations you have in your data.
+Instead of choosing the number of bins, 
+we suggest you choose the bin width that makes sense for your problem.
+This can be set via the `binwidth` argument in `geom_histogram`.
+
+Choosing the correct bin width for your problem 
+is something that commonly takes iteration.
+It's usually a good idea to try out several bin widths.
+Let's do that here for the histogram we have been working on.
+In Figure \@ref(fig:03-data-morley-hist-binwidth)
```
leem44, line 856:
```suggestion
In Figure \@ref(fig:03-data-morley-hist-binwidth),
```

viz.Rmd
```diff
@@ -569,23 +820,115 @@ for just how accurate all the experiments were overall. For example, how accurat
...
+we compare the default number of bins with three histograms where we set the 
+`binwidth` in `geom_histogram` to 0.001, 0.01 and 0.1.
+We can see that in this case, both the default number of bins
```
leem44, line 859:
```suggestion
In this case, we can see that both the default number of bins
```

viz.Rmd
```diff
@@ -569,23 +820,115 @@ for just how accurate all the experiments were overall. For example, how accurat
...
+
+
+
+#### Adding layers to a `ggplot` plot object {-}
+
+What if we want to add onto a `ggplot` plot object later in our analysis?
+Do we need to copy and paste the code we wrote previously 
+and then add on our new layers after that? No! 
+If we stored our `ggplot` plot as a named object
+using the assignment symbol (`<-`) then we can call that object
```
leem44, line 911:
```suggestion
using the assignment symbol (`<-`), we can call it 
```

viz.Rmd
```diff
@@ -569,23 +820,115 @@ for just how accurate all the experiments were overall. For example, how accurat
...
+
+
+#### Adding layers to a `ggplot` plot object {-}
+
+What if we want to add onto a `ggplot` plot object later in our analysis?
+Do we need to copy and paste the code we wrote previously 
+and then add on our new layers after that? No! 
+If we stored our `ggplot` plot as a named object
+using the assignment symbol (`<-`) then we can call that object 
+and add on to it using the `+` operator.
```
leem44, line 912:
```suggestion
and add to it using the `+` operator.
```

viz.Rmd
```diff
@@ -569,23 +820,115 @@ for just how accurate all the experiments were overall. For example, how accurat
...
+> the take home message to the audience. 
+> Typically, 
+> that is the answer to the question you posed before making the visualization.
 
 ## Explaining the visualization
-#### *Tell a story*
+#### *Tell a story* {-}
 
 Typically, your visualization will not be shown entirely on its own, but rather it will be part of a larger presentation.
 Further, visualizations can provide supporting information for any aspect of a presentation, from opening to conclusion.
```
leem44, line 934: space between colon and "story" in this line: "Regardless of where it appears, a good way to discuss your visualization is as a story :"

leem44, line 934: below it says "distributions (scatters, histograms)" -- shouldn't this just be "distributions (histograms)"

viz.Rmd
```diff
@@ -610,12 +953,12 @@ Regardless of where it appears, a good way to discuss your visualization is as a
...
-more heat from the sun, leading to global warming. (2) In order to assess how quickly the atmospheric concentration of CO2 is increasing over time, we (3) used a data set from
-the Mauna Loa observatory from Hawaii, consisting of CO2 measurements from 1959 to the present. We plotted the measured concentration of CO2 (on the vertical axis) over time
+**Mauna Loa Atmospheric CO$_{\text{2}}$ Measurements:** (1) \index{Mauna Loa CO2} Many current forms of energy generation and conversion---from automotive engines to natural gas power plants---rely on burning
+fossil fuels and produce greenhouse gases, typically primarily carbon dioxide (CO$_{\text{2}}$), as a byproduct. Too much of these gases in the Earth's atmosphere will cause it to trap
+more heat from the sun, leading to global warming. (2) In order to assess how quickly the atmospheric concentration of CO$_{\text{2}}$ is increasing over time, we (3) used a data set from
+the Mauna Loa observatory from Hawaii, consisting of CO$_{\text{2}}$ measurements from 1959 to the present. We plotted the measured concentration of CO$_{\text{2}}$ (on the vertical axis) over time
```
leem44, line 959: this says we used measurements from 1959 to present but I think this was changed right? 1980s -2020 or something?

viz.Rmd
```diff
@@ -671,35 +1012,67 @@ vector graphics as much as you like without the image looking bad, while raster
...
+       `Image size` = c(paste(round(file.info("img/faithful_plot.png")["size"] / 1000000, 2), "MB"),
+                        paste(round(file.info("img/faithful_plot.jpg")["size"] / 1000000, 2), "MB"),
+                        paste(round(file.info("img/faithful_plot.bmp")["size"] / 1000000, 2), "MB"),
+                        paste(round(file.info("img/faithful_plot.tiff")["size"] / 1000000, 2), "MB"),
+                        paste(round(file.info("img/faithful_plot.svg")["size"] / 1000000, 2), "MB")))
+kable(file_sizes,
+      caption = "File sizes of `faithful_plot` when saved as different file formats.")
 ```
+
+The different file types are not only different in their properties
```
leem44, line 1067: formatting in this section (writing for the rest of the chapter centered rather than left aligned and the caption formatting displays in the pdf) 
Screen Shot 2021-10-01 at 1 31 48 PM

viz.Rmd
```diff
@@ -671,35 +1012,67 @@ vector graphics as much as you like without the image looking bad, while raster
...
+The different file types are not only different in their properties 
+(outlined above), but also in their file sizes (Table \@ref(tab:filesizes)).
 Wow, that's quite a difference! Notice that for such a simple plot with few
 graphical elements (points), the vector graphics format (SVG) is over 100 times
 smaller than the uncompressed raster images (BMP, TIFF). Also, note that the
 JPG format is twice as large as the PNG format since the JPG compression
-algorithm is designed for natural images (not plots). Below, we also show what
+algorithm is designed for natural images (not plots). 
+
+Below, we also show what
```
leem44, line 1075: images are stacked in the pdf instead of side by side -- they also look the same... 
Screen Shot 2021-10-01 at 1 33 23 PM

viz.Rmd
```diff
@@ -716,3 +1089,12 @@ knitr::include_graphics("img/faithful_zoom.svg")
 - The [`ggplot2` page on the tidyverse website](https://ggplot2.tidyverse.org) is where you should look if you want to learn more about the functions in this chapter, the full set of arguments you can use, and other related functions. The site also provides a very nice cheat sheet that summarizes many of the data wrangling functions from this chapter.
```
leem44, line 1089: should provide the links here -- I guess this will be a part of the formatting pass

leem44 (commented):
A couple of things that I didn't comment on but we should add to the formatting pass: 
- sometimes the caption has periods at the end, sometimes it doesn't -- we should add this to the style guide which one we want 
- the plots and code coming off the page + the placement of figures -- add to formatting pass

viz.Rmd
```diff
@@ -36,13 +36,22 @@ By the end of the chapter, readers will be able to:
...
 ## Choosing the visualization
-#### *Ask a question, and answer it*
+#### *Ask a question, and answer it* {-}
 
 The purpose of a visualization is to answer a question \index{question!visualization} about a data set of interest. So naturally, the first thing to do **before** creating a visualization is to formulate the question about the data you are trying to answer. 
 A good visualization will clearly answer your question without distraction; a *great* visualization will suggest even what the question
 was itself without additional explanation. Imagine your visualization as part of a poster presentation for a project; even if you aren't standing
 at the poster explaining things, an effective visualization will convey your message to the audience.
 
+Tables are another medium used to answer to questions with data
```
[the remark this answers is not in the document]
ttimbers: We are actually going to remove this section, so will not accept this suggestion.

viz.Rmd
```diff
@@ -36,13 +36,22 @@ By the end of the chapter, readers will be able to:
...
 
 The purpose of a visualization is to answer a question \index{question!visualization} about a data set of interest. So naturally, the first thing to do **before** creating a visualization is to formulate the question about the data you are trying to answer. 
 A good visualization will clearly answer your question without distraction; a *great* visualization will suggest even what the question
 was itself without additional explanation. Imagine your visualization as part of a poster presentation for a project; even if you aren't standing
 at the poster explaining things, an effective visualization will convey your message to the audience.
 
+Tables are another medium used to answer to questions with data 
+and communicate the answers.
+However, visualizations are usually more effective 
+at communicating the big picture, take home message.
```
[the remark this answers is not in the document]
ttimbers: We are actually going to remove this section, so will not accept this suggestion.

viz.Rmd
```diff
@@ -36,13 +36,22 @@ By the end of the chapter, readers will be able to:
...
+and communicate the answers.
+However, visualizations are usually more effective 
+at communicating the big picture, take home message. 
+For this reason, data visualizations are widely used 
+and the focus of this chapter.
+However, when the goal of communication is precision and details, 
+tables can then be a better choice.
+
 Recall [the different data analysis questions](index.html#chapter-learning-objectives) from the first chapter. With the visualizations
 we will cover in this chapter, we will be able to answer *only descriptive and exploratory* questions. Be careful not to try to answer any *predictive, inferential, causal or mechanistic*
```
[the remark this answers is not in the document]
ttimbers: I am not sure the suggested rewording keeps the intended meaning. I think @trevorcampbell was trying to say to not answer predictive, inferential, causal or mechanistic questions with viz, not just be careful. I will try to reword.

viz.Rmd
ttimbers: Addressed the order. I do think we could move `canlang` to being after co2 data. But eruption is not at all about a scatter plot, that is about saving data. So we should leave that one at the end.

ttimbers: Duh, I am silly, the eruption data is there too. Yes, I agree, let's move islands to after canlang!

viz.Rmd
```diff
@@ -176,47 +275,123 @@ To add axis labels, we use the `xlab` and `ylab` functions. To change the font s
...
   ylab("Atmospheric CO2 (ppm)") +
-  scale_x_continuous(limits = c(1990, 1995)) +
-  theme(text = element_text(size = 18))
+  xlim(c(date("1990-01-01"), date("1994-12-01"))) +
+  theme(text = element_text(size = 16))
 co2_line
 ```
 
-Interesting! It seems that each year, the atmospheric CO2 increases until it reaches its peak somewhere around April, decreases until around late September, 
+Interesting! It seems that each year, the atmospheric CO$_{\text{2}}$ increases until it reaches its peak somewhere around April, decreases until around late September,
```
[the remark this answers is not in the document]
ttimbers: I changed it so we only look at 4 years, then we get a grid line at 6 months (June) so it's a bit easier to follow now.

viz.Rmd
```diff
@@ -176,47 +275,123 @@ To add axis labels, we use the `xlab` and `ylab` functions. To change the font s
...
 islands_df <- read_csv("data/islands.csv")
 islands_df
 ```
-Here, we have a list of Earth's landmasses, and are trying to compare their sizes. The right type of visualization to answer this
-question is a bar plot, specified by the `geom_bar` function in `ggplot2`. However, by default, `geom_bar` sets the heights
+
+Here, we have a data frame of Earth's landmasses, 
+and are trying to compare their sizes. 
+The right type of visualization to answer this question is a bar plot. 
+A bar plot uses a rectangle to encode the value a summary statistic
```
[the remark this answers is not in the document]
ttimbers: addressed.

viz.Rmd
```diff
@@ -238,48 +413,68 @@ islands_bar <- ggplot(islands_top12, aes(x = size, y = landmass)) +
...
+but by size, and to colour them based on whether they are a continent. 
+The data for this is stored in the `landmass_type` column. 
+To use this to colour the bars, 
+we add the `fill` argument to the aesthetic mapping 
+and set it to `landmass_type`. 
+
+To organize the landmasses by their `size` variable, 
+we will use the `tidyverse` `fct_reorder` function
+in the aesthetic mapping to organize the landmasses by their `size` variable.
+The first argument passed to `fct_reorder` is the factor columns's name,
```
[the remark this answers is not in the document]
ttimbers: addressed.

viz.Rmd
```diff
@@ -238,48 +413,68 @@ islands_bar <- ggplot(islands_top12, aes(x = size, y = landmass)) +
...
+The data for this is stored in the `landmass_type` column. 
+To use this to colour the bars, 
+we add the `fill` argument to the aesthetic mapping 
+and set it to `landmass_type`. 
+
+To organize the landmasses by their `size` variable, 
+we will use the `tidyverse` `fct_reorder` function
+in the aesthetic mapping to organize the landmasses by their `size` variable.
+The first argument passed to `fct_reorder` is the factor columns's name, 
+whose level's we would like to reorder (here `landmass`).
```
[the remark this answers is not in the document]
ttimbers: addressed.

viz.Rmd
```diff
@@ -238,48 +413,68 @@ islands_bar <- ggplot(islands_top12, aes(x = size, y = landmass)) +
...
+we will use the `tidyverse` `fct_reorder` function
+in the aesthetic mapping to organize the landmasses by their `size` variable.
+The first argument passed to `fct_reorder` is the factor columns's name, 
+whose level's we would like to reorder (here `landmass`). 
+The second argument passed is the column name 
+that holds the values we would like to use to do the ordering (here `size`).
+`fct_reorder` orders by that column in ascending order by default, 
+but this can be changed to sorting in descending order 
+by setting  `.desc = TRUE`.
+We do this here so that the largest bar will closet to the axis line,
```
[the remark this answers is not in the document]
ttimbers: fixed

viz.Rmd
```diff
@@ -238,48 +413,68 @@ islands_bar <- ggplot(islands_top12, aes(x = size, y = landmass)) +
...
+as that is more visually appealing.
```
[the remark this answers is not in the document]
ttimbers: fixed

viz.Rmd
```diff
@@ -238,48 +413,68 @@ islands_bar <- ggplot(islands_top12, aes(x = size, y = landmass)) +
...
+
+To label the x and y axes, we shift from using `xlab` and `ylab`, 
+to specifying these within the `labs` function. 
+We do this here because we would also like to change the legend label.
+The default label is the column name of the column being mapped to `fill`,
+however `landmass_type` is not very human readable or visually pleasing.
```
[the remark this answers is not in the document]
ttimbers: fixed

viz.Rmd
```diff
@@ -238,48 +413,68 @@ islands_bar <- ggplot(islands_top12, aes(x = size, y = landmass)) +
...
+### The Old Faithful eruption/waiting time data set with scatter plots
+The `faithful` data set \index{Old Faithful} contains measurements 
+of the waiting time between eruptions 
+and the subsequent eruption duration (in minutes). 
+The `faithful` data set is available in base R as a data frame,
+so it does not need to be loaded.
+We convert it to a tibble to take advantage of the nicer print output 
+these specialized data frames provide.
 
 **Question:** \index{question!visualization} Is there a relationship between the waiting time before an eruption to the duration of the eruption?
```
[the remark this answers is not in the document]
ttimbers: fixed

viz.Rmd
```diff
@@ -238,48 +413,68 @@ islands_bar <- ggplot(islands_top12, aes(x = size, y = landmass)) +
...
-```{r 03-data-faithful-tibble, echo = F, warning=FALSE, message=FALSE}
-# old faithful eruption time / wait time data convert to tibble
-faithful <- as.tibble(faithful)
-```
-
 ```{r 03-data-faithful, warning=FALSE, message=FALSE}
 # old faithful eruption time / wait time data
+faithful <- as_tibble(faithful)
 faithful
 ```
```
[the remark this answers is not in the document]
ttimbers: I think what is trying to be said is that they do not have a natural order - like time does.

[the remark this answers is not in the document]
ttimbers: More formally, the difference is that the time variable is the CO2 data is not independent, whereas here both variables are are.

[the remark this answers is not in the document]
ttimbers: @leem44 @trevorcampbell I have changed it to this to make it clearer (I want to avoid the term independent, as we have not introduced it, nor do we elsewhere in the book):

```
Here again, we investigate the relationship between two quantitative variables 
(waiting time and eruption time). 
But if you look at the output of the data frame, 
you'll notice that unlike time in the Mauna Loa CO$_{\text{2}}$ data set,
neither of the variables here have a natural order to them.
So, in this case, let's start again with a scatter plot:
```

viz.Rmd
```diff
@@ -343,12 +538,14 @@ can_lang |>
 To answer our question, we will need to adjust the scale of the x and y axes so that they are on a log scale. Doing this can help make patterns in the data more interpretable.
```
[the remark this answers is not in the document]
ttimbers: done

viz.Rmd
```diff
@@ -343,12 +538,14 @@ can_lang |>
...
-  ylab("Mother tongue \n (number of Canadian residents)") +
-  scale_x_log10(labels = scales::comma) +
-  scale_y_log10(labels = scales::comma)
+  labs(x = "Language spoken most at home \n (number of Canadian residents)",
+       y = "Mother tongue \n (number of Canadian residents)") +
+  scale_x_log10(labels = comma) +
+  scale_y_log10(labels = comma)
 ```
 > Here we are adjusting the scale of the x and y axes on a $log$ scale with base 10. 
 > Notice that $log_{10}(10) = 1$, $log_{10}(100) = 2$, and $log_{10}(1000) = 3$.
```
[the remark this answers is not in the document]
ttimbers: @leem44 - I tried this, for relationship direction, strength and linearity. What do you think?

```
When we see two variables moving in the same direction 
(as x increases, y increases), we call this a *positive relationship*.
If instead, we saw two variables moving in the opposite direction 
(as x increases, y decreases), we would call this a *negative relationship*.
Another possibility here, is to observe no relationship. 
This is when y neither nor decreases across the values of x.
Because the points are fairly close together, 
we can say that the relationship is strong. 
If instead, we saw that the points were spread much further apart, 
we would say that the relationship was weak.
Because drawing a straight line through these points 
would fit the pattern we observe quite well, we say that it's linear.
If instead, drawing a straight line through these points 
did not fit the observed pattern of the data, 
we would say that it is non-linear.
```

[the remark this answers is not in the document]
ttimbers: I added the word context here and refer to chapter 3 where we discuss this in more detail.

viz.Rmd
```diff
@@ -404,32 +601,34 @@ We can add an argument to the `aes` function, specifying that the `category` col
...
+  scale_y_log10(labels = comma)
 ```
 
-In Figure \@ref(fig:03-scatter-colour-by-category), the points are coloured with the default `ggplot2` colour palette. But what if you want to use different colours? In R, one package that provides alternative colour palettes \index{color palette} is `RColorBrewer` [@RColorBrewer]. You can visualize the list of colour palettes that `RColorBrewer` has to offer with the `display.brewer.all()` function. You can also print a list of colour blind friendly palettes by adding `colorblindFriendly = T` to the function.
+In Figure \@ref(fig:03-scatter-colour-by-category), the points are coloured with the default `ggplot2` colour palette. But what if you want to use different colours? In R, one package that provides alternative colour palettes \index{color palette} is `RColorBrewer` [@RColorBrewer]. You can visualize the list of colour palettes that `RColorBrewer` has to offer with the `display.brewer.all` function. You can also print a list of colour blind friendly palettes by adding `colorblindFriendly = TRUE` to the function.
 
 ```{r}
 library(RColorBrewer)
-display.brewer.all(colorblindFriendly = T)
+display.brewer.all(colorblindFriendly = TRUE)
```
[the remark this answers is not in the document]
ttimbers: done

viz.Rmd
```diff
@@ -404,32 +601,34 @@ We can add an argument to the `aes` function, specifying that the `category` col
...
+To change the colour palette, we add the `scale_colour_brewer` layer indicating the palette we want to use. 
+
 ```{r scatter-colour-by-category-palette, fig.width=7.75, fig.height=4, warning=FALSE,  fig.cap = "Scatter plot of percentage of Canadians reporting a language as their mother tongue vs the primary language at home coloured by language category with colour blind friendly colours"}
 ggplot(can_lang, aes(x = most_at_home_percent, y = mother_tongue_percent, color = category)) +
   geom_point() +
-  xlab("Language spoken most at home \n (percentage of Canadian residents)") +
-  ylab("Mother tongue \n (percentage of Canadian residents)") +
-  scale_x_log10(labels = scales::comma) +
-  scale_y_log10(labels = scales::comma) + 
+  labs(x = "Language spoken most at home \n (number of Canadian residents)",
```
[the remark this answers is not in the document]
ttimbers: done

viz.Rmd
```diff
@@ -468,34 +667,61 @@ ggplot(can_lang, aes(
   shape = category
 )) + # map categories to different shapes
   geom_point(alpha = 0.6) + # set the transparency of the points
-  xlab("Language spoken most at home \n (percentage of Canadian residents)") +
-  ylab("Mother tongue \n (percentage of Canadian residents)") +
-  scale_x_log10(labels = scales::comma) +
-  scale_y_log10(labels = scales::comma) + 
+  labs(x = "Language spoken most at home \n (number of Canadian residents)",
+       y = "Mother tongue \n (number of Canadian residents)") +
```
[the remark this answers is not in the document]
ttimbers: done

viz.Rmd
```diff
@@ -404,32 +601,34 @@ We can add an argument to the `aes` function, specifying that the `category` col
...
+  scale_x_log10(labels = comma) +
+  scale_y_log10(labels = comma) +
   scale_color_brewer(palette = "RdYlBu")
 ```
 
 You can use this [color blindness simulator](https://www.color-blindness.com/coblis-color-blindness-simulator/) \index{color palette!color blindness simulator} to check if your visualizations are colour blind friendly.
```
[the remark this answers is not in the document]
ttimbers: Will deal with in formatting pass.

viz.Rmd
```diff
@@ -389,10 +586,10 @@ Finally, let's visualize the data now that we have represented it as percentages
 ```{r 03-mother-tongue-vs-most-at-home-scale-props, fig.width=5.75, fig.height=5, warning=FALSE, fig.cap = "Scatter plot of percentage of Canadians reporting a language as their mother tongue vs the primary language at home"}
 ggplot(can_lang, aes(x = most_at_home_percent, y = mother_tongue_percent)) +
   geom_point() +
-  xlab("Language spoken most at home \n (percentage of Canadian residents)") +
-  ylab("Mother tongue \n (percentage of Canadian residents)") +
-  scale_x_log10(labels = scales::comma) +
-  scale_y_log10(labels = scales::comma)
+  labs(x = "Language spoken most at home \n (number of Canadian residents)",
```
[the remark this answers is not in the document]
ttimbers: fixed

viz.Rmd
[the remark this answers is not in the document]
leem44: missing word: This is when y neither **increases** nor decreases...
I think we should add to the sentence: "Because the points in our graph are fairly close together," -- since we are going from talking about the general case to this specific case

viz.Rmd
```diff
@@ -328,8 +523,8 @@ To improve the interpretability of the data visualization in Figure \@ref(fig:03
 ```{r 03-mother-tongue-vs-most-at-home-labs, fig.width=5.75, fig.height=5, warning=FALSE, fig.cap = "Scatter plot of number of Canadians reporting a language as their mother tongue vs the primary language at home with x and y labels"}
```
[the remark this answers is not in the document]
ttimbers: It is currently rendering in the html, will check again in the PDF when I do the global formatting pass.

viz.Rmd
```diff
@@ -389,10 +586,10 @@ Finally, let's visualize the data now that we have represented it as percentages
...
-  ylab("Mother tongue \n (percentage of Canadian residents)") +
-  scale_x_log10(labels = scales::comma) +
-  scale_y_log10(labels = scales::comma)
+  labs(x = "Language spoken most at home \n (number of Canadian residents)",
+       y = "Mother tongue \n (number of Canadian residents)") +
+  scale_x_log10(labels = comma) +
+  scale_y_log10(labels = comma)
 ```
 
 Now we'll move onto the second part of our exploratory data analysis question: when considering the relationship between the number of people who have a language as their mother tongue and the number of people who speak that language at home, is there a pattern in the strength of this relationship in the higher-level language categories (Official languages, Aboriginal languages, or non-official and non-Aboriginal languages)? One common way to explore this is to colour the data points on the scatter plot we have already created by group. For example, given that we have the higher-level language category for each language recorded in the 2016 Canadian census, we can colour the points in our previous
```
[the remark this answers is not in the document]
ttimbers: fixed

viz.Rmd
```diff
@@ -404,32 +601,34 @@ We can add an argument to the `aes` function, specifying that the `category` col
...
+       y = "Mother tongue \n (number of Canadian residents)") +
+  scale_x_log10(labels = comma) +
+  scale_y_log10(labels = comma) +
   scale_color_brewer(palette = "RdYlBu")
 ```
 
 You can use this [color blindness simulator](https://www.color-blindness.com/coblis-color-blindness-simulator/) \index{color palette!color blindness simulator} to check if your visualizations are colour blind friendly.
+Additionally, the `ggthemes` R package also has a [color blind friendly color pallete](https://jrnold.github.io/ggthemes/reference/colorblind.html).
 
 From the visualization in Figure \@ref(fig:scatter-colour-by-category-palette), we can now clearly see that not just a lot, but that the majority of Canadians reported one of the official languages as their mother tongue and as the language they speak most often at home.
```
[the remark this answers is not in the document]
ttimbers: fixed

viz.Rmd
```diff
@@ -468,34 +667,61 @@ ggplot(can_lang, aes(
   shape = category
 )) + # map categories to different shapes
```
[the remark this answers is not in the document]
ttimbers: fixed

viz.Rmd
```diff
@@ -468,34 +667,61 @@ ggplot(can_lang, aes(
...
 
-### The Michelson speed of light data set
-The `morley` data set \index{Michelson speed of light} contains measurements of the speed of light (in kilometres per second with 299,000 subtracted) from the year 1879 
-for five experiments, each with 20 consecutive runs [@lightdata]. This data set is available in base R under the name `morley` so it does not need to be loaded.
+### The Michelson speed of light data set with histograms
+The `morley` data set \index{Michelson speed of light} 
+contains measurements of the speed of light 
+collected in experiments performed in 1879.
+Five experiments were performed, 
+and in each experiment 20 runs were performed -
```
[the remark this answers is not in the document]
ttimbers: done

viz.Rmd
```diff
@@ -468,34 +667,61 @@ ggplot(can_lang, aes(
...
 
 ```{r 03-data-morley, warning=FALSE, message=FALSE}
 # michelson morley experimental data
+morley <- as_tibble(morley)
 morley
 ```
 
 In this experimental data, Michelson was trying to measure just a single quantitative number (the speed of light). The data set
 contains many measurements of this single quantity. To tell how accurate the experiments were, we need to visualize the 
 distribution of the measurements (i.e., all their possible values and how often each occurs). We can do this using a *histogram*. A histogram \index{ggplot!histogram} helps us visualize how a
```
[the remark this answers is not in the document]
ttimbers: We do have it at the beginning of the chapter. But good idea adding it to the index.

viz.Rmd
```diff
@@ -468,34 +667,61 @@ ggplot(can_lang, aes(
...
 ```{r 03-data-morley, warning=FALSE, message=FALSE}
 # michelson morley experimental data
+morley <- as_tibble(morley)
 morley
 ```
 
 In this experimental data, Michelson was trying to measure just a single quantitative number (the speed of light). The data set
 contains many measurements of this single quantity. To tell how accurate the experiments were, we need to visualize the 
 distribution of the measurements (i.e., all their possible values and how often each occurs). We can do this using a *histogram*. A histogram \index{ggplot!histogram} helps us visualize how a 
 particular variable is distributed in a data set by separating the data into bins, and then using vertical bars
```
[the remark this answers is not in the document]
ttimbers: But we can create histograms for non-quantitative data, so I would rather leave it as "particular".

viz.Rmd
```diff
@@ -504,15 +730,30 @@ morley_hist <- ggplot(morley, aes(x = Speed)) +
...
+unless we can see what the true value is.
+In order to visualize the true speed of light, 
+we will add a vertical line with the `geom_vline` function 
+\index{ggplot!geom\_vline}.
+To draw a vertical line with `geom_vline`, 
+we need to specify where on the x-axis the line should be drawn. 
+We can do this by setting the `xintercept` argument. 
+Here we set it to the true value of light speed minus 299, 000 (so it is coded the same way as the measurements in the `morley` data frame): 792.458. 
+We also would like to fine tune this vertical line, 
+styling it so that is dashed and 1 point in thickness.
```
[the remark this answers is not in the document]
ttimbers: I equivocated it to what it is in mm. And added it is a measurement unit we commonly also use with font.

viz.Rmd
ttimbers: We explain it in the landmass example. We don't even really need it here at all... I don't think...

ttimbers: OH WAIT - that is stat = "identity" in the landmass example... Let me add the explanation for `position = "identity" ` here.

ttimbers: And we kind of need it in the one plot where we try to map the colours and fail - because otherwise, the shape changes and that is confusing.

viz.Rmd
ttimbers: We can't really without going down a deep rabbit hole about unquoted column names in R, and how to program with them... 

For example, here are the docs for `vars`:
Just like `aes()`, `vars()` is a quoting function that takes inputs to be evaluated in the context of a dataset. 

To explain that we need to explain why R is weird by using unquoted column names, and what it means to evaluate something in the context of a dataset - which means explaining the global environment... and on...

viz.Rmd
ttimbers: Rug plots, and 1D scatters can be useful for 1D distributions in some scenarios:

Rug plot:
Screen Shot 2021-10-01 at 11 33 19 PM

1D scatter with 3 groups:
Screen Shot 2021-10-01 at 11 33 49 PM

ttimbers: I guess we don't really teach those here, so if you feel strongly, we can remove.

viz.Rmd
ttimbers: OK, this is weird and should not happen, I will check tables in the formatting pass.

viz.Rmd
ttimbers: Yep, aware and will fix when we attack the PDF

viz.Rmd
ttimbers: yep

viz.Rmd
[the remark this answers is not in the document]
leem44: My edit is wrong, its still missing the word "be": "the largest bar will **be** closest"

ttimbers:
All feedback from reviewers, @trevorcampbell and @leem44 addressed. Merging! 🚀

PR #268, opened 2021-10-03
added total WSSD plot and fixed a bug in the plots.

ttimbers:


clustering.Rmd
```diff
@@ -295,12 +303,81 @@ for (i in 1:nrow(clus1)) {
...
+             size = 5)
 base
 ```
 
 The larger the value of $S^2$, the more spread-out the cluster is, since large $S^2$ means that points are far from the cluster center.
-Note, however, that "large" is relative to *both* the scale of the variables for clustering *and* the number of points in the cluster. A cluster where points are very close to the center might still have a large $S^2$ if there are many data points in the cluster. 
+Note, however, that "large" is relative to *both* the scale of the variables for clustering *and* the number of points in the cluster. A cluster where points are very close to the center might still have a large $S^2$ if there are many data points in the cluster.
+
+After we have calculated the WSSD for all the clusters, 
+we sum them together to to get the *total WSSD*.
```
trevorcampbell, line 318: "to to"

clustering.Rmd
```diff
@@ -295,12 +303,81 @@ for (i in 1:nrow(clus1)) {
...
+For our example, 
+this means adding up all the distances of the black lines in 
+Figure \@ref(fig:10-toy-example-all-clus-dists).
```
trevorcampbell, line 321: distances? or squared distances?

clustering.Rmd
```diff
@@ -295,12 +303,81 @@ for (i in 1:nrow(clus1)) {
...
+
+```{r 10-toy-example-all-clus-dists, echo = FALSE, warning = FALSE, fig.height = 4, fig.width = 5, fig.cap = "All clusters from the `penguin_data` data set example. Observations are in orange blue, and yellow with the cluster center highlighted in red. The distances from the observations to each of the respective cluster centers are represented as black lines."}
```
trevorcampbell, line 323: "orange blue"

PR #274, opened 2021-10-03
changing to american spelling, fixing minor spelling/grammar errors, …

leem44:
…changing out.width in plots to percent instead of pt, chaning underscores to dashes in code blocks


ttimbers:
Thanks for this @leem44 - merging into `pdf` and then we can do one PR into `dev`.

PR #384, opened 2021-12-02
preamble copyedit pass

leem44:
Made changes to preface, acknowledgements and authors files


trevorcampbell:
@leem44 `&mdash;` is an em-dash, I believe the typesetter asked for an en-dash `&ndash;`. I'm going to make this change and merge.

PR #387, opened 2021-12-03
wrangling edit pass 

leem44:
pg 107 she changed it to "right-hand" but I think it should be "left-hand" still? please double check

pg 110 heading "Calculating summary statistics when there are NAs" she wrote: "Please be sure that this heading level, unnumbered, is as you intend."

I think this is supposed to be the sublevel of 3.9.1 but do we want to change this? Should we just rename this section to "Calculating summary statistics on whole columns when there are NAs" and make it section 3.9.2?

### DIFF
--- a/wrangling.Rmd
+++ b/wrangling.Rmd
@@ -19,14 +19,14 @@ application, providing more practice working through a whole case study.
 
 ## Chapter learning objectives
 
-By the end of the chapter, readers will be able to:
+By the end of the chapter, readers will be able to do the following:
 
-  - define the term "tidy data"
-  - discuss the advantages of storing data in a tidy data format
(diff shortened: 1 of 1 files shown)


trevorcampbell:
> pg 107 she changed it to "right-hand" but I think it should be "left-hand" still? please double check

You are correct

> pg 110 heading "Calculating summary statistics when there are NAs" she wrote: "Please be sure that this heading level, unnumbered, is as you intend."

> I think this is supposed to be the sublevel of 3.9.1 but do we want to change this? Should we just rename this section to "Calculating summary statistics on whole columns when there are NAs" and make it section 3.9.2?

I like this idea. Will edit + merge.

Issue #402, opened 2021-12-08
Check footnotes


trevorcampbell:
there are some odd ones that don't work, e.g. in the current draft on page 51.

We need to do a global pass on this.

trevorcampbell:
also from R Alexander:

> There are only two pages of citations. While possible this really is all
> the material that the authors draw on, it might be that some R
> packages are inadvertently not being cited, and that the decision to
> include links to books in footnotes, rather than as citations (e.g. p. 122
> but other places also) is artificially reducing the number of citations.

trevorcampbell:
- p2: the trc.ca link no longer works
- p41: oracle link doesnt go to DB, it goes to oracle company
- p50 and 59: twitter link needs https, it currently links to local file
- p51: broken url
- p60: link has parentheses
- p126: broken url

done through chp 4, still working

trevorcampbell:
#409 closed this

PR #403, opened 2021-12-08
Rohan Alexander edits

### LINKED ISSUE #379: Feedback to address in v1 from @RohanAlexander
- [x] There are only two pages of citations. While possible this really is all 
the material that the authors draw on, it might be that some R 
packages are inadvertently not being cited, and that the decision to 
include links to books in footnotes, rather than as citations (e.g. p. 122 
but other places also) is artificially reducing the number of citations.
- [x] The phrase ‘scientific disciplines’ or ‘applications’ is used on p. 277 and
p. 327 seemed unusual to me and I didn’t understand what ‘scientific’ 
was meant to refer to?
- [x] p. 25: `filter` 
- [x] p. 48: Odd spacing between ‘.csv’ and ‘file’.
- [x] p. 51 (and elsewhere in footnotes): Odd links that don’t work.
- [x] p. 59: ‘chapter 3’, instead of ‘Chapter 3’.
- [x] p. 65 ‘Tidyverse twitter’ instead of ‘Tidyverse Twitter’.
- [x] Generally, some inconsistency about ‘tidyverse’ or ‘Tidyverse’ throughout.
- [x] p. 109: ‘the data set the no’ instead of ‘the data set that no’.
- [x] p. 110: ‘see that no longer’ instead of ‘we see that we no longer’.
- [x] p. 114: missing paragraph spacing before ‘Note that...’?
- [x] p. 175: odd spacing in the two bullet points at the end.
- [x] p. 194: overfull line extending over the edge of the page.
- [x] p. 303: ‘help make our’ instead of ‘help make sure our’

trevorcampbell:
Closes #379

and also changing a lot of "our ...." to "the ...." -- less awk


trevorcampbell:
@leem44 i opened another issue for that since it's a global pass thing! #402

Issue #372, opened 2021-10-27
Increase plot font sizes after chapter 4


ttimbers:
So we are pretty picky about font sizes in the viz chapter, and then in the classification chapter we just use the default font size. We should eventually change this. Ideally, we set a default font size in the setup chunk at the beginning. I think this is possible.

leem44:
addressed in #399

leem44:
addressed in #405

PR #405, opened 2021-12-08
Plot font sizes

leem44:
This PR addresses #372 
- Increased plot font sizes in global knitr chunk for all chapters after viz 
- also added `theme(text = element_text(size = 14))` for any code that is displayed so it matches the figure output


trevorcampbell:
@leem44 

- axis labels still small in section 6.8.1 - 6.8.3

(will continue adding comments here as I go through)

Issue #406, opened 2021-12-10
Missing software from installation chapter


ttimbers:
### R packages to add:
- ~~`testthat`~~ (was there already, we just missed it)
- ~~`cowplot`~~ (was there already, we just missed it)
- `scales`

### R packages to remove:
- `infer` (it is now installed via `tidymodels`)

PR #407, opened 2021-12-10
remove infer and add scales

ttimbers:
- scales is used in textbook and worksheets
- infer is now installed by tidymodels


ttimbers:
Closes #406

PR #413, opened 2022-01-08
Patch versions setup

### LINKED ISSUE #408: specific r package versions in setup chapter
We should edit the install instructions in the setup chapter to include the specific set of package versions that we know work together to run our worksheets.

ttimbers:
Closes #408

### DIFF
--- a/setup.Rmd
+++ b/setup.Rmd
@@ -121,38 +121,43 @@ tab in your default browser with the JupyterLab interface. To exit out of
 JupyterLab you can click `File -> Shutdown`, or go to the terminal from which 
 you launched JupyterLab, hold `Ctrl`, and press `C` twice.
 
-### R, R packages and the IRkernel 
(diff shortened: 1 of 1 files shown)


trevorcampbell (approved):
Looks good to me (I made some minor changes throughout, have a look and then merge when you're ready).

trevorcampbell:
- some minor polish
- changed "our..." and "we..." to "your..." and "you..."  (these are instructions)

@ttimbers I think the instructions for installing miniconda are probably fairly brittle. It's a specific set of vague commands like "press enter", "type yes", etc. So if miniconda ever changes the order/prompts during installation, these will break.

Perhaps it's just better to give broader advice here instead of specific actions? Like "follow the prompts and agree to blah, make sure you tell it to run `conda init`, etc"

ttimbers:
Good point @trevorcampbell - will remove those and replace with a more general statement and then merge this.

PR #416, opened 2022-01-09
Patch whitespace

ttimbers:


ttimbers:
No changes seem to be needed for regression 1 🙌

trevorcampbell:
Just reviewed your work up to & including clustering.

HTML: LGTM!

PDF:
- p.29 and p.282 (just combining into one comment): the beginning of a note box is at the bottom of the page here. Can we just add like one newline (latex compile only) to push this onto the next page?
- p.34 (content): Just noticed a minor bug -- It says `\t` is an escape character -- no, `\` is the escape character, and `\t` is an escaped `t` used to denote a tab. We should fix this. For example, we can write
  ```
  **Note:** `\t` is an example of an *escaped character*, which always starts with a backslash (`\`). Escaped characters are used to 
  represent non-printing characters (like the tab) or characters with special meanings (such as quotation marks).
  ```
- page 228, 229 (content): Figures 6.9, 6.10, 6.11 need thicker lines (maybe just 1 pt thicker). It's weird because figure 6.6 has thicker lines but doesn't do anything special to achieve that. maybe some hidden code running there.

ttimbers:
No changes seem to be needed for jupyter  🙌

trevorcampbell:
Just reviewed Inference + Jupyter -- both look fine to me!

trevorcampbell:
All looks good to me!

Issue #422, opened 2022-02-03
Final comments from Michele


trevorcampbell:
- add a blank page after dedication
- page numbers at the beginning (fixing #314 should cause 2-compile latex and solve this, I think?)
- in the acknowledgments, capitalize the word "Foreword"
- in Figures 1.7 and 1.8, I suggest we write in the caption something like "This figure is not quite done yet; there are still many improvements to be made!"
- p51 : "Stop! can you scrape..." should be shifted to the next page (latex spacing... for this one we need to be careful that it doesn't mess up anything else afterwards in that chapter)
- p241 sentence structure in note box
- p378 backwards quote
- p412 bad typesetting in index
- end of chapter blurbs #421

trevorcampbell:
Handled in #423

Issue #421, opened 2022-01-16
Change verbiage and link in exercise section


ttimbers:
Currently the exercise section links directly to the worksheet on GitHub, but skips over the landing page where they can use binder. We were going to change this, but missed doing this. Here's what it currently says:

"Exercises
Practice exercises for the material covered in this chapter can be found in the accompanying worksheet. The worksheet tries to provide automated feedback and help guide you through the problems. To make sure this functionality works as intended, please follow the instructions for computer setup needed to run the worksheets found in Chapter 13."

I think we wanted to add a link to the table with the worksheets and Binderhub links to each of these....

trevorcampbell:
Handled in #423

Issue #314, opened 2021-10-08
index "data science definition" is on the wrong page 


leem44:

leem44:
the page number in the index says the wrong page

trevorcampbell:
This can probably be fixed by moving the `\index{...}` command closer to the actual word. I occasionally put it nearby but perhaps not close enough when I went through and made the index.

trevorcampbell:
It's a bug in tinytex -- see the bookdown repo issue -- I've documented where the problem is.

I think the best solution for us (unless the rstudio folks move quickly) is to edit our dockerfile to install a hacked version of tinytex that runs the latex build twice before makeindex.

trevorcampbell:
The `tinytex` devs actually handled this immediately. But to avoid depending on an unstable release of tinytex, I still suggest we hack our own version of it explicitly in the dockerfile when we install 0.34

trevorcampbell:
Handled in #423

PR #423, opened 2022-02-03
Copyediting round 2

trevorcampbell:
This PR closes:
- #422 (various fixes)
- #314  (fixed by forcing `tinytex` to use the current master branch in the repo, which has the fix for index 2x compile. I also had to use the latest `xfun`, which `tinytex` seems to depend on.)
- #421 (edited the blurb and link to point to the worksheets repo readme)

Note 0: the edits should be self-explanatory except one in `reading.Rmd`, where it looks like I just changed a few words for no reason. That was to force the "**Stop! Are you allowed....**" onto the next page.

Note 1: right now the container is at `v0.22.0` on `master` (and dockerhub), which fixes the index hyperref issue. It's also `v0.22.0` in the build scripts on this branch.

But because I originally edited the dockerfile on this branch (before I realized I had to do it on master), I eventually made the same edit directly on master and merged into this branch. So I suspect when we merge this PR into dev and then master, it might bump the docker container again. In that case you're safe to edit `build_pdf.sh` and `build_html.sh` to use container `v0.23.0` -- nothing will change between `v0.22.0` and `v0.23.0`.

Note 2: The docker image would not build unless I specified an updated version of the `cancensus` package (which `canlang` depends on). I had to pin cancensus to a git commit hash. So **please double-check that nothing broke anywhere we use `canlang`.**

Note 3: I was unable to get a single blank page between the dedications and the TOC. I could only get either 2 or 0. I believe LaTeX forces 2 blank pages here because the TOC has to start on an odd page. **We should notify Michele of this, but I don't think we need to do anything ourselves**

Note 4: You'll notice in the Exercises blurbs that I point to https://github.com/UBC-DSCI/data-science-a-first-intro-worksheets#readme (instead of https://ubc-dsci.github.io/data-science-a-first-intro-worksheets/ ). The "view worksheets" links don't work on the github.io site; they download the worksheet instead of previewing it in github. I also fixed the link and blurb text at the end of the preface where we mention worksheets.

Note 5: We should also tell Michele directly that Fig 1.7 and 1.8 are meant to have broken x axis labels (in addition to me editing the caption), because that was also flagged in the first copyediting pass.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -39,6 +39,7 @@ RUN Rscript -e "reticulate::install_miniconda()"
 RUN Rscript -e "reticulate::conda_install('r-reticulate', 'python-kaleido')"
 RUN Rscript -e "reticulate::conda_install('r-reticulate', 'plotly', channel = 'plotly')"
 
+RUN Rscript -e "devtools::install_github('mountainMath/cancensus@5a5d61759d477986d40dd87fa9a6532ff6037efe')"
 RUN Rscript -e "devtools::install_github('ttimbers/canlang@0.0.1')"
 
 # install LaTeX packages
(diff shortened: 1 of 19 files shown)


leem44:
@trevorcampbell Thank you for making these changes! It looks good to me!

PR #433, opened 2022-04-19
Ch5-8

leem44:
- fixing spacing between words, pages etc 
- minor changes to wording

### DIFF
--- a/classification2.Rmd
+++ b/classification2.Rmd
@@ -134,35 +134,34 @@ Here, we pass in the number `1`.
 
 ```{r}
 set.seed(1)
-random_numbers <- sample(0:9, 10, replace=TRUE)
-random_numbers
+random_numbers1 <- sample(0:9, 10, replace=TRUE)
+random_numbers1
 ```
 
-You can see that `random_numbers` is a list of 10 numbers
+You can see that `random_numbers1` is a list of 10 numbers
 from 0 to 9 that, from all appearances, looks random. If 
 we run the `sample` function again, we will 
 get a fresh batch of 10 numbers that also look random.
 
 ```{r}
-random_numbers <- sample(0:9, 10, replace=TRUE)
-random_numbers
+random_numbers2 <- sample(0:9, 10, replace=TRUE)
+random_numbers2
 ```
 
 If we want to force R to produce the same sequences of random numbers,
 we can simply call the `set.seed` function again with the same argument
-value.
+value. 
 
 ```{r}
 set.seed(1)
-random_numbers <- sample(0:9, 10, replace=TRUE)
-random_numbers
+random_numbers1_again <- sample(0:9, 10, replace=TRUE)
+random_numbers1_again
 
-set.seed(1)
-random_numbers <- sample(0:9, 10, replace=TRUE)
-random_numbers
+random_numbers2_again <- sample(0:9, 10, replace=TRUE)
+random_numbers2_again
 ```
 
-And if we choose 
+Notice that after setting the seed, we get the same two sequences of numbers in the same order. `random_numbers1` and `random_numbers1_again` produce the same sequence of numbers, and the same can be said about `random_numbers2` and `random_numbers2_again`. And if we choose 
 a different value for the seed&mdash;say, 4235&mdash;we
 obtain a different sequence of random numbers.
 
@@ -171,7 +170,6 @@ set.seed(4235)
 random_numbers <- sample(0:9, 10, replace=TRUE)
 random_numbers
 
-set.seed(4235)
 random_numbers <- sample(0:9, 10, replace=TRUE)
 random_numbers
 ```
--- a/classification2.Rmd
+++ b/classification2.Rmd
@@ -120,7 +120,7 @@ in the analysis, would we not get a different result each time?
 The trick is that in R&mdash;and other programming languages&mdash;randomness 
 is not actually random! Instead, R uses a *random number generator* that
 produces a sequence of numbers that
-are completely determined by a \index{seed} \index{random seed|see{seed}}
+are completely determined by a\index{seed} \index{random seed|see{seed}}
  *seed value*. Once you set the seed value 
 using the \index{seed!set.seed} `set.seed` function, everything after that point may *look* random,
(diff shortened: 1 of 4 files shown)


regression2.Rmd
```diff
@@ -129,7 +129,7 @@ $\beta_1$ as the increase in price for each square foot of space.
 Let's push this thought even further: what would happen in the equation for the line if you 
 tried to evaluate the price of a house with size 6 *million* square feet?
 Or what about *negative* 2,000 square feet? As it turns out, nothing in the formula breaks; linear
-regression will happily make predictions for crazy predictor values if you ask it to. But even though
+regression will happily make predictions for different predictor values if you ask it to. But even though
```
trevorcampbell, line 132: "different" doesn't convey the right sentiment here -- instead of crazy, how about "nonsensical" ?

classification2.Rmd
```diff
@@ -157,6 +157,7 @@ set.seed(1)
 random_numbers <- sample(0:9, 10, replace=TRUE)
 random_numbers
 
+set.seed(1)
```
trevorcampbell, line 160: I don't think this is correct... why add this?

classification2.Rmd
```diff
@@ -170,6 +171,7 @@ set.seed(4235)
 random_numbers <- sample(0:9, 10, replace=TRUE)
 random_numbers
 
+set.seed(4235)
```
trevorcampbell, line 174: I don't think this is correct -- why add this?

regression1.Rmd
```diff
@@ -747,7 +745,6 @@ sacr_multi <- workflow() |>
 
 sacr_k <- sacr_multi |>
               pull(neighbors)
-
```
trevorcampbell, line 750: Not sure I know why spaces are being deleted here -- any particular reason?

regression2.Rmd
```diff
@@ -553,8 +553,8 @@ $$\text{house sale price} = \beta_0 + \beta_1\cdot(\text{house size}) + \beta_2\
 where:
 
 - $\beta_0$ is the *vertical intercept* of the hyperplane (the price when both house size and number of bedrooms are 0)
-- $\beta_1$ is the *slope* for the first predictor (how quickly the price increases as you increase house size)
-- $\beta_2$ is the *slope* for the second predictor (how quickly the price increases as you increase the number of bedrooms)
+- $\beta_1$ is the *slope* for the first predictor (how quickly the price changes as you increase house size holding number of bedrooms constant)
```
trevorcampbell, line 556: need comma

regression2.Rmd
```diff
@@ -553,8 +553,8 @@ $$\text{house sale price} = \beta_0 + \beta_1\cdot(\text{house size}) + \beta_2\
...
+- $\beta_2$ is the *slope* for the second predictor (how quickly the price changes as you increase the number of bedrooms holding house size constant)
```
trevorcampbell, line 557: need comma

regression2.Rmd
leem44: Yeah "nonsensical" is a better word! Will change!

classification2.Rmd
leem44: So this is how the output looks without the change:
Screen Shot 2022-04-19 at 11 42 08 AM

So we get a different set of random numbers -- I imagine the intention is that we want to show the same?

classification2.Rmd
leem44: This is the original version: 
Screen Shot 2022-04-19 at 11 42 39 AM

So we are getting different numbers

leem44: Or is that what you wanted to show? Perhaps I misunderstood the paragraph

regression1.Rmd
leem44: For spacing of the code blocks spilling onto new pages

classification2.Rmd
trevorcampbell: maybe the text wasn't clear -- the intention was to show that if you call sample twice (after set seed), you get the same two sequences of numbers in the same order. 

I would prefer not to call set seed twice, because then students might be tricked into thinking you have to set the seed before every function call that you want to be reproducible. 

happy to entertain more clear wording choice here!

classification2.Rmd
trevorcampbell: same comment as above

regression1.Rmd
trevorcampbell: nope, just wasn't sure why the deletion happened! if it looks good to you, then i'm happy with it

classification2.Rmd
leem44: Oh, I see! Yeah okay, that wasn't clear to me, I thought you were trying to show that set.seed(1) would give the same set of numbers twice in a row. The first one is showing a set.seed. Then the second random_numbers is showing a different set. Then I thought the last two were a side by side comparison of the set.seed(1) to show the same sequence of numbers.  Shall we change the wording then? 
perhaps if we assign the variables as "random_numbers1", "random_numbers2", "random_numbers3" and "random_numbers4" 
then we can reference the set of numbers and say: 
"Notice that after setting the seed, we get the same two sequences of numbers in the same order. `random_numbers1` and `random_numbers3` produce the same sequence of numbers after setting the seed. The same can be said about `random_numbers2` and random_numbers4`."

trevorcampbell: yup that works. but maybe make the link even clearer with names like `random_numbers1` and `random_numers1_again`
and `random_numbers2`, `random_numbers2_again`

leem44: Perfect! Thanks!

regression2.Rmd
```diff
@@ -706,8 +706,8 @@ sacramento_train <- sacramento_train |>
 
 lm_plot_multicol_1 <- ggplot(sacramento_train) +
   geom_point(aes(x = sqft, y = sqft1), alpha=0.4)+
-  xlab("House size (square feet)") +
-  ylab("House size (square inches)") 
+  xlab("House size 1 (square feet)") +
```
trevorcampbell, line 709: "House size measurement 1 / House size measurement 2" 

maybe a bit clearer?

PR #435, opened 2022-05-05
Ch1 4

ttimbers:
Pull request for proof edits for chapters 1-4. Work in progress (chapters 1 & 2) are done except for this issue for chapter 2:

warning line should come before the Note... but it's not for some reason... (page 33 of book, source is in `reading.Rmd`)

Screen Shot 2022-05-04 at 8 20 09 PM

(same issue is present in html version of book...)

Source: 

Screen Shot 2022-05-04 at 8 21 28 PM

This appears to be a `bookdown` thing... because if instead I use the knit button (or `rmarkdown::render`) it does what you would expect:

Screen Shot 2022-05-04 at 8 30 37 PM


ttimbers:
OK @trevorcampbell - I have now added all the edits I view as "essential" and "scoped" to the proof. I still however don't have a fix for the one I highlighted above... Otherwise this is ready for your review.

reading.Rmd
```diff
@@ -1,3 +1,8 @@
+---
+output:
+  pdf_document: default
```
trevorcampbell, line 3: What's this stuff?

reading.Rmd
```diff
@@ -169,8 +174,11 @@ We put `data/` before the file's
 name when we are loading the data set because this data set is located in a
 sub-folder, named `data`, relative to where we are running our R code.
 
-Here is what the file would look like in a plain text editor (a program that removes
-all formatting, like bolding or different fonts):
+Here is what the file would look like in a plain text editor.
+
+> Note: A plain text editor is a program that that allows for editing of plain text. 
+Plain text is text represented only by characters
```
trevorcampbell, line 180: I found this note box to be a bit abstract/unclear. What is "meta-specification data", "character specification data", "paragraph specification data", etc ?

reading.Rmd
ttimbers: Ohhh, let's remove that. I must have got put there when I click "knit" to test the note/warning stuff above. I didn't notice RStudio did that...

reading.Rmd
ttimbers: That is the formatting data from rich text. We can change how this is worded, but we cannot simply revert back to what we had before because it was wrong (plain text editors do no "remove" formatting from text/rich text. They just represent everything as characters.

reading.Rmd
```diff
@@ -1,3 +1,8 @@
+---
+output:
+  pdf_document: default
+  html_document: default
+---
```
ttimbers, line 5:
```suggestion
```

reading.Rmd
trevorcampbell: removed

ttimbers:
Unaddressed issue at top of thread opened here: https://github.com/UBC-DSCI/introduction-to-datascience/issues/436

PR #439, opened 2022-06-13
Copyright patch

ttimbers:
Looking at our repo for this book, it is not clear from the repo who the original authors are in the README, and who owns the copyright the license is distributing the material under, as well as how to attribute us. We should fix this.

The print version of the book says: "© 2022 Tiffany Timbers, Trevor Campbell and Melissa Lee" so I think we still retain copyright of all materials. So this PR attempts to address this and tries to make it more clear in our book source repo.


trevorcampbell:
LGTM, merging

PR #442, opened 2022-06-21
Patch bootstrap img

ttimbers:
Fix to source image and jpeg. Fix does the following:
- replaces 3 (not in original sample) with 9 (in original sample)
- replaces mean of 2.5 (calculated with 3 in bootstrap sample) with mean of 3.5 (calculated with 9 in bootstrap sample)


trevorcampbell:
@ttimbers did you forget to re-commit the source file? Based on the diffs it looks like the source file was just deleted.

trevorcampbell (approved):
Looks good to me

PR #448, opened 2022-09-22
Correct parameter name in `facet_grid`

joelostblom:
The parameter used for faceting into columns is called `cols` rather than `columns` https://ggplot2.tidyverse.org/reference/facet_grid.html. I noticed this when going through the slides for DSCI-100.


trevorcampbell:
Great catch -- thank you!

PR #486, opened 2023-07-06
Minor edits summer 2023

### LINKED ISSUE #437: some writing needs work in 2.4.3
```
Let’s compare the data frame here to the resulting data frame in Section 2.4.1
after using read_csv . Notice anything? They look the same! The same number
of columns/rows and column names! So we needed to use different tools for
the job depending on the file format and our resulting table ( canlang_data ) in
both cases was the same!
```

Bit too many exclamation points in a row, and sentence fragments

### LINKED ISSUE #456: switch to `-` instead of `+` for example of broken variable name in Ch1
See 

https://github.com/UBC-DSCI/introduction-to-datascience-python/issues/53

which also applies here. We should fix both at the same time

### LINKED ISSUE #457: minor wording in chapter 1
"Greatest to least" -> "largest to smallest value"

### LINKED ISSUE #459: minor wording chapter 2
> Notice anything? They look the same! The same number of columns/rows and column names! So we needed to use different tools for the job depending on the file format and our resulting table (canlang_data) in both cases was the same!

This is a bit awkward -- as we are working through the python book, I rephrased this:

> 
>  Notice anything? They look the same; they have
> the same number of columns and rows, and have the same column names! 
> So even though we needed to use different
> arguments depending on the file format, our resulting data frame
> (`canlang_data`) in both cases was the same.

### LINKED ISSUE #463: minor wording chapter 5
"Based on our visualization, it seems like the prediction of an unobserved label might be possible." weird sentence"

changed in py version to

"Based on our visualization, it seems like it may be possible to make accurate predictions of the Class variable (i.e., a diagnosis) for tumor images with unknown diagnoses."

### LINKED ISSUE #466: Minor error in Chapter 12
The `add_collab_03.png` figure is missing its explanation in the text.

It should say

> Then click the green "Invite a collaborator" button.

### LINKED ISSUE #467: Better URL for worksheets
The exercises at the end of each chapter point to the github repo. They should point instead to the rendered website

https://ubc-dsci.github.io/data-science-a-first-intro-worksheets

It would be especially nice to configure a subdomain of datasciencebook.ca to be for the worksheets.

### LINKED ISSUE #471: Minor wording Ch6
Rewrite this "We can evaluate a classifier by splitting the data randomly into a training and test data set, using the training set to build the classifier, and using the test set to estimate its accuracy. Finally, we can tune the classifier (e.g., select the number of neighbors in K-NN) by maximizing estimated accuracy via cross-validation." to this "We can tune and evaluate a classifier by splitting the data randomly into a training and test data set. The training set is used to build the classifier and we can tune the classifier (e.g., select the number of neighbors in K-NN) by maximizing estimated accuracy via cross-validation. After we have tuned the model we can use the test set to estimate its accuracy.

### LINKED ISSUE #472: Error in axis labels Chapter 5 last figure
The axes say "standardized", but these plots were made using non-standardized data (the pipeline standardizes before classification). Should check in the surrounding text/code to make sure removing the "(standardized)" from axis labels won't break anything else.

### LINKED ISSUE #476: Weird linebreak in inference (but only for small windows)
In the bootstrapping section, where the text reads

```
Let’s now take 20,000 bootstrap samples from the original sample (one_sample)
using rep_sample_n, and calculate the means for each of those replicates.
```

If you shrink the window horizontally, it'll break weirdly. There's probably some extra whitespace in the `.md` file at the end of the line.

### LINKED ISSUE #485: can_lang.tsv vs can_lang_tab.tsv
In reading chapter: bit odd that `_tab` is used for the file with headings, and no `_tab` is for without headings. `tab` seems to stand for "tabs as delimiters", which is already specified by `.tsv`.

trevorcampbell:
This addresses a whole lot of small things that have built up over time. The project detailing what was done here is https://github.com/orgs/UBC-DSCI/projects/5/views/1 (in progress are items in this PR -- they will automatically move to "done" once we merge.)
- closes #476 
- closes #472 
- closes #471 
- closes #467 
- closes #466 
- closes #459 
- closes #463 
- closes #457 
- closes #456 
- closes #437 
- closes #485

### DIFF
--- a/data/can_lang.tsv
+++ b/data/can_lang.tsv
@@ -1,3 +1,4 @@
+category	language	mother_tongue	most_at_home	most_at_work	lang_known
 Aboriginal languages	Aboriginal languages, n.o.s.	590	235	30	665
 Non-Official & Non-Aboriginal languages	Afrikaans	10260	4785	85	23415
 Non-Official & Non-Aboriginal languages	Afro-Asiatic languages, n.i.e.	1150	445	10	2775
--- a/data/can_lang_no_names.tsv
+++ b/data/can_lang_no_names.tsv
@@ -1,4 +1,3 @@
-category	language	mother_tongue	most_at_home	most_at_work	lang_known
 Aboriginal languages	Aboriginal languages, n.o.s.	590	235	30	665
 Non-Official & Non-Aboriginal languages	Afrikaans	10260	4785	85	23415
 Non-Official & Non-Aboriginal languages	Afro-Asiatic languages, n.i.e.	1150	445	10	2775
--- a/source/classification1.Rmd
+++ b/source/classification1.Rmd
@@ -213,7 +213,8 @@ Before we start doing any modeling, let's explore our data set. Below we use
 the `group_by`, `summarize` and `n` \index{group\_by}\index{summarize} functions to find the number and percentage 
 of benign and malignant tumor observations in our data set. The `n` function within
 `summarize`, when paired with `group_by`, counts the number of observations in each `Class` group. 
-Then we calculate the percentage in each group by dividing by the total number of observations. We have 357 (63\%) benign and 212 (37\%) malignant tumor observations.
+Then we calculate the percentage in each group by dividing by the total number of observations
+and multiplying by 100. We have 357 (63\%) benign and 212 (37\%) malignant tumor observations.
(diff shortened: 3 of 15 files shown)


chendaniely (approved):
Took a quick glance through all the changes in the files. Nothing looks out of place

PR #509, opened 2023-07-13
Missing data

### LINKED ISSUE #11: Missing data
-  Additionally, I need to open another issue about things to add to the course. In the projects students encountered a lot of missing data, and we didn’t formally discuss what to do with such tings in this book… So we should at least discuss some aspect of this here (not fully cover imputation for example, but instead discuss options such as removing observations, and doing simple mean imputation perhaps, and when to do which).

trevorcampbell:
closes #11 

(there is a certain satisfaction that comes with making a PR to close issue 11 in a repo with nearly 500 issues and 2000+ commits)

( :D )

### DIFF
--- a/source/classification1.Rmd
+++ b/source/classification1.Rmd
@@ -1294,11 +1294,91 @@ upsampled_plot <-
 upsampled_plot
 ```
 
+### Missing data
+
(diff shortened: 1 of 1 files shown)


trevorcampbell:
Reviewed by @leem44 , who approved via Slack. Merging!

PR #514, opened 2023-08-05
Added mutate to ch1

### LINKED ISSUE #512: Add mutate into Chapter 1
If we merge https://github.com/UBC-DSCI/introduction-to-datascience-python/pull/188 in the Python book, we should make the same edit here with `mutate`. The rationale there is that `mutate`/`assign` is used in the Ch1 and 2 worksheets/tutorials, so it makes sense to introduce that in the book earlier so that students see it prior to encountering it in the assignments.

trevorcampbell:
Closes #512 

- makes Ch1 more consistent with Py
- resolves issue where `mutate` is in Ch1,2 worksheets but not textbook

### DIFF
--- a/source/intro.Rmd
+++ b/source/intro.Rmd
@@ -445,7 +445,7 @@ selected_lang <- select(aboriginal_lang, language, mother_tongue)
 selected_lang
 ```
 
-### Using `arrange` to order and `slice` to select rows by index number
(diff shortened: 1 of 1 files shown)


trevorcampbell:
(this was previously created by @joelostblom and reviewed by myself in the Py version in https://github.com/UBC-DSCI/introduction-to-datascience-python/pull/188 -- merging here).

PR #517, opened 2023-08-05
Refreshing setup material

### LINKED ISSUE #446: Install instruction bugs
Two bugs came up in our install instructions, reported by an external person (running windows). Here are their comments:

- when running

`conda env update --file https://raw.githubusercontent.com/UBC-DSCI/data-science-a-first-intro-worksheets/main/environment.yml`

unixodbc=2.3.9 was not found 

image000

-  tinytex won’t install, I just saw that the error messages are shown in Anaconda and not in the R console. Tinytex only installed after I downloaded the tinytex library, which was not part of the instructions.

`install.packages(“tinytex”)`

We should try to find a way to fix these such that the instructions that are in the book still work properly

### LINKED ISSUE #484: Setup chapter doesn't create reproducible environment for worksheets
We've had at least a few external people try to use our worksheets in an environment that doesn't line up with our docker image, resulting in frustration when tests don't pass. We should consider re-writing the setup chapter to focus on getting our environment set up (as opposed to general package installation etc)

Also do this for the python version

trevorcampbell:
This PR introduces install and run instructions for Docker Desktop and JupyterLab Desktop. A bit simplified compared with what we had previously, and more likely to produce a consistent environment for working on the worksheets. Also far more likely to be doable by our students in DSCI100, which means we may be able to offload some computation from our centralized server. Also removes unnecessary stuff like LaTeX install.

Closes #484 
Closes #446 

~~**Reviewers:** Please try out these instructions and let me know if they don't work. I do not have a MacOS computer, so could not test that out.~~ we've tested it all now. all good


trevorcampbell:
Also @joelostblom given our discussion [here](https://github.ubc.ca/UBC-DSCI/dsci-100-instructor/issues/610#issuecomment-22783) I wonder if you could streamline the JupyterLab installation by somehow providing a pre-baked environment file that gets set up during the install, as you mentioned?

source/setup.Rmd
```diff
@@ -1,282 +1,288 @@
...
 
 > You can also install the *latest* version of R 
-> and the R packages used in this book by typing the commands shown below 
-> in the Anaconda Prompt (Windows) 
-> or terminal (MacOS and Ubuntu) and pressing enter.
+> and the R packages used in this book by instead typing the commands shown below 
 > **Be careful though:** this may install package versions that are
 > incompatible with the worksheets that accompany the book; the automated
 > exercise feedback might tell you your answers are not correct even though
 > they are!
```
joelostblom, line 237: Could we fix the package versions to what's in the docker container to avoid this?

source/setup.Rmd
```diff
@@ -1,282 +1,288 @@
...
+- If you need to enable virtualization, you will need to edit your BIOS. Restart your computer, and enter the BIOS using the hotkey
+  (usually Delete, Esc, and/or one of the F# keys). Look for an "Advanced" menu, and under your CPU settings, set the "Virtualization" option
+  to "enabled". Then save the changes and reboot your machine. If you are not familiar with BIOS editing, you may want to find an expert 
+  to help you with this, as editing the BIOS can be dangerous. Detailed instructions for doing this are beyond the scope of this book. 
+More information on installing and running
+Docker on Windows, as well as troubleshooting tips, can be found [in the online
+Docker documetation](https://docs.docker.com/desktop/install/windows-install/).
+
+
+Once Docker is running (in either Windows or MacOS), you need to download and run the virtual machine that we have made available
```
joelostblom, line 133: Maybe this could have its own header to make it clearer that it belongs to windows and mac?

joelostblom (commented):
Just reading through the changes, this looks good to me. I don't have time to work through the setup myself at the moment, and I don't have access to a Mac either so maybe someone else with a Mac can go through the instructions to test them?

source/setup.Rmd
trevorcampbell: yeah i thought about doing this, but was feeling lazy. Since you also pointed it out I'll make sure to do this (both for JLab desktop and Docker, which need a similar edit)

source/setup.Rmd
trevorcampbell: So, the `environment.yml` file is the one people should use to get precisely the right versions of things. I might actually just remove this note box to avoid the issue entirely.

source/setup.Rmd
trevorcampbell: done!

trevorcampbell:
For readers of this PR: the Mac instructions have now been tested on `MacOS 13 Ventura`. We are waiting to check 
- docker and jupyterlab desktop on an intel Mac computer
- jupyterlab desktop on a windows computer

Assuming both are OK, we should then be able to merge.

source/setup.Rmd
```diff
@@ -1,282 +1,277 @@
...
+this setup manually can be quite tricky, as it involves quite a few different
+software packages, not to mention getting the right versions of
+everything&mdash;the worksheets and autograder tests may not work unless all the versions are
+exactly right! To keep things simple, we instead recommend that you install
+[Docker](https://docker.com). Docker lets you run your Jupyter notebooks inside
+a pre-built *container* that comes with precisely the right versions of 
+all software packages needed run the worksheets that come with this book.
+\index{Docker}
+
+### Ubuntu
```
ttimbers, line 66: This might seem trivial, but perhaps we should list the operating systems in order of likelihood of being used by the students? Most students run Windows, followed by Mac, and only a few run Linux. Having a word, like "Ubuntu" early on that many students do not recognize might be intimidating and discourage some from following the instructions.

source/setup.Rmd
```diff
@@ -1,282 +1,277 @@
...
+sudo apt install ca-certificates curl gnupg
+curl -fsSL https://get.docker.com -o get-docker.sh
+sudo chmod u+x get-docker.sh
+sudo sh get-docker.sh
 ```
 
-**Ubuntu:** To install Git on Ubuntu, open the terminal 
-and type the following commands:
-
+**Running JupyterLab** In the terminal, navigate to the directory where you want to run JupyterLab, and run
```
ttimbers, line 77: Do we assume students know how to do this?

source/setup.Rmd
```diff
@@ -1,282 +1,277 @@
...
+download the `Docker.dmg` installation file that is appropriate for your
+computer. To know which installer is right for your machine, you need to know 
+whether your computer has an Intel processor (older machines) or an
+Apple processor (newer machines); the [Apple support page](https://support.apple.com/en-ca/HT211814) has
+information to help you determine which processor you have. Once downloaded, double-click 
+the file to open the installer, then drag the Docker icon to the Applications folder.
+Double-click the icon in the Applications folder to start Docker. In the installation 
+window, use the recommended settings.
+
+**Running JupyterLab** Once Docker is installed and running, follow the
```
ttimbers, line 106: This section also leads me to believe we should have the Windows section first, given that we are telling students to scroll down to that section and follow the instructions there?

source/setup.Rmd
```diff
@@ -1,282 +1,277 @@
...
+- If the admin account on your computer is different to your user account, you must add the user to the "docker-users" group. 
+  Run Computer Management as an administrator and navigate to `Local Users` and `Groups -> Groups -> docker-users`. Right-click to 
+  add the user to the group. Log out and log back in for the changes to take effect.
+- If you need to enable virtualization, you will need to edit your BIOS. Restart your computer, and enter the BIOS using the hotkey
+  (usually Delete, Esc, and/or one of the F# keys). Look for an "Advanced" menu, and under your CPU settings, set the "Virtualization" option
+  to "enabled". Then save the changes and reboot your machine. If you are not familiar with BIOS editing, you may want to find an expert 
+  to help you with this, as editing the BIOS can be dangerous. Detailed instructions for doing this are beyond the scope of this book. 
+More information on installing and running
+Docker on Windows, as well as troubleshooting tips, can be found [in the online
+Docker documentation](https://docs.docker.com/desktop/install/windows-install/).
```
ttimbers, line 130: I think this should be presented as a note, as it's only needed if something goes wrong.

source/setup.Rmd
```diff
@@ -1,282 +1,277 @@
...
-the default installation location,
-and to running `conda init`, which makes `conda` available from the terminal.
+## Working with JupyterLab Desktop
 
-**Ubuntu:** To install miniconda on Ubuntu, first download
-the [latest Python 64-bit version from here](https://repo.anaconda.com/miniconda/Miniconda3-latest-Linux-x86_64.sh). 
-After the download has finished, open the terminal and execute the following 
-command:
+You can also run the worksheets accompanying this book on your computer
+using [JupyterLab Desktop](https://github.com/jupyterlab/jupyterlab-desktop).
```
ttimbers, line 176: I make this suggestion as they are still "using" JupyterLab in Docker, the bigger difference here is that they are installing JupyterLab directly onto their OS.
```suggestion
by installing [JupyterLab Desktop](https://github.com/jupyterlab/jupyterlab-desktop).
```

source/setup.Rmd
```diff
@@ -1,282 +1,277 @@
...
+Docker can sometimes run into some fairly technical issues (especially on Windows computers) 
+that require expert troubleshooting. The downside of JupyterLab Desktop is that there is a (very) small chance that 
+you may not end up with the right versions of all the R packages needed for the worksheets. Docker, on the other hand,
+*guarantees* that the worksheets will work exactly as intended. 
 
-```
-bash path/to/Miniconda3-latest-Linux-x86_64.sh
-```
+In this section, we will cover how to install Git (for version control, as discussed in Chapter \@ref(Getting-started-with-version-control)),
+JupyterLab Desktop, the JupyterLab Git extension, and all of the R packages needed to run the code in this book.
```
ttimbers, line 184: I suggest this change because in the paragraph above we are focused on the JupyterLab software, and then we make a hard transition to saying we are going to install Git which we have not yet talked about in this chapter...
```suggestion
In this section, we will cover how to install JupyterLab Desktop, Git (for version control, as discussed in Chapter \@ref(Getting-started-with-version-control)), the JupyterLab Git extension, and all of the R packages needed to run the code in this book.
```

source/setup.Rmd
```diff
@@ -1,282 +1,277 @@
...
+\index{git!installation}\index{JupyterLab Desktop}
 
-Make sure to replace `path/to/` with the path of the folder containing the downloaded
-script. Most often this file will be downloaded to the `Downloads` folder.
-If this is the case for your computer, you can run the script in the terminal by typing:
+### Ubuntu
```
ttimbers, line 187: Again, I think we should go Windows, Mac then Ubuntu...

source/setup.Rmd
```diff
@@ -1,282 +1,277 @@
...
+Apple processor (newer machines); the [Apple support page](https://support.apple.com/en-ca/HT211814) has
+information to help you determine which processor you have.
+Once downloaded, double-click the file to open the installer, then drag
+the JupyterLab Desktop icon to the Applications folder.  Double-click
+the icon in the Applications folder to start JupyterLab Desktop.
+
+**Configuring JupyterLab Desktop** From this point onward, with JupyterLab Desktop running,
+follow the instructions in the Ubuntu section on *Configuring JupyterLab Desktop* to set up the
+environment, install the JupyterLab Git extension, and install
+the various R software packages needed for the worksheets.
```
ttimbers, line 276: I think organzation-wise, it might be best to have a section called "All operating systems"? For common steps? It is hard to refer back to where exactly the students should go to, in particular if we are not using headers that can be linked to.

source/setup.Rmd
```diff
@@ -1,282 +1,277 @@
...
+the software required to work on Jupyter notebooks on your own computer. Doing
+this setup manually can be quite tricky, as it involves quite a few different
+software packages, not to mention getting the right versions of
+everything&mdash;the worksheets and autograder tests may not work unless all the versions are
+exactly right! To keep things simple, we instead recommend that you install
+[Docker](https://docker.com). Docker lets you run your Jupyter notebooks inside
+a pre-built *container* that comes with precisely the right versions of 
+all software packages needed run the worksheets that come with this book.
+\index{Docker}
+
```
ttimbers, line 65: Should we have a note here about what a container is? something like:

>Note: In software, a **container** is a virtualized operating system user space. This allows software to be run in isolation, and not interfere with the software that is installed in your operating system's user space. This means that you can use a container to run a specific version of the R programming language (and the necessary packages) for the book's worksheets even if you have a different version of R installed on your computer in your operating system user space. Or even if you haven't installed R at all!

source/setup.Rmd
```diff
@@ -1,282 +1,277 @@
...
+  add the user to the group. Log out and log back in for the changes to take effect.
+- If you need to enable virtualization, you will need to edit your BIOS. Restart your computer, and enter the BIOS using the hotkey
+  (usually Delete, Esc, and/or one of the F# keys). Look for an "Advanced" menu, and under your CPU settings, set the "Virtualization" option
+  to "enabled". Then save the changes and reboot your machine. If you are not familiar with BIOS editing, you may want to find an expert 
+  to help you with this, as editing the BIOS can be dangerous. Detailed instructions for doing this are beyond the scope of this book. 
+More information on installing and running
+Docker on Windows, as well as troubleshooting tips, can be found [in the online
+Docker documentation](https://docs.docker.com/desktop/install/windows-install/).
+
+**Running JupyterLab** Once Docker is running (in either Windows or MacOS), you need to download and run the
```
ttimbers, line 132: Do we need a sentence "Open Docker Desktop"?

source/setup.Rmd
```diff
@@ -1,282 +1,277 @@
...
+```{r docker-desktop-images, echo = FALSE, fig.cap = "The Docker Desktop images tab.", fig.retina = 2, out.width="100%"}
+image_read("img/setup/docker-2.png") |> 
+  image_crop("3632x2000")
+```
 
-If your Mac computer has an Apple M1 processor chip you can download 
-the [latest Python 64-bit version from here](https://repo.anaconda.com/miniconda/Miniconda3-latest-MacOSX-arm64.sh).
-After the download has finished, you need to run the downloaded script in the terminal using a command
-like:
+To start up a *container* using that image, click the play button beside the image. This will open the run configuration menu (Figure \@ref(fig:docker-desktop-runconfig)).
```
ttimbers, line 153: Note: on a Mac I had to also click on the down arrow/carrot to see the configuration menu (on Mac it is called "Optional settings"

ttimbers: Screenshot 2023-08-28 at 12 13 06 PM

Screenshot 2023-08-28 at 12 13 14 PM

source/setup.Rmd
```diff
@@ -1,282 +1,277 @@
...
+In the "Host port" textbox, enter `8888`. In the "Volumes" section, click the "Host path" box and navigate to the folder where your Jupyter worksheets
+are stored. In the "Container path" text box, enter `/home/jovyan/work`. Then click the "Run" button to start the container.
```
ttimbers, line 155: We really should give some recommendations on the following:

1. Where student should store the downloaded worksheets (I would suggest documents)
2. What path to suggest for the Host path. I know in older versions of Docker, Docker has major performance issues if the host path is set to the computers home directory on mac. It essentially becomes unusable... So we want to avoid students from doing that...

source/setup.Rmd
```diff
@@ -1,282 +1,277 @@
...
-If this is the case for your computer, you can run the script in the terminal by typing:
+After clicking the "Run" button, you will see a terminal. The terminal will then print 
+some text as the Docker container starts. Once the text stops scrolling, find the 
+URL in the terminal that starts 
+with `http://127.0.0.1:8888` (highlighted by the red box in Figure \@ref(fig:ubuntu-docker-terminal)), and paste it 
+into your browser to start JupyterLab.
 
-```
-bash Downloads/Miniconda3-latest-MacOSX-arm64.sh
+```{r docker-desktop-url, echo = FALSE, fig.cap = "The terminal text after running the Docker container in Ubuntu. The red box indicates the URL that you should paste into your browser to open JupyterLab.", fig.retina = 2, out.width="100%"}
```
ttimbers, line 168: Why are we showing what the terminal looks like in Ubuntu under the Windows instructions? This might be confusing for students.

source/setup.Rmd
ttimbers: We also need to provide some shutdown instructions if we want students to use Docker Desktop on Mac and Windows. This is because when you run a container with the UI (as opposed to the command line) it depends how you stop it as to whether you get the behaviour you want that is equivalent to  to `--rm`.  To get the equivalent to `--rm` students have to click the red, not the stop button or the grey garbage can, it has to be the red garbage can.

Screenshot 2023-08-28 at 12 31 35 PM

If they click the stop button, the container it doesn't get deleted and instead still sits there stopped, but using the port allocated to it. This means no new container can take the port until that one is deleted. Alternatively students could just restart the container via Docker Desktop to get up and running again, but we need to teach them how to do that.

source/setup.Rmd
trevorcampbell: I thought this as well, but I remember at the time finding that there was a reason it was harder to do that way in terms of flow. It's far enough in the future now that I don't remember, so I'll give it a shot.

source/setup.Rmd
trevorcampbell: Good idea, I will change that

source/setup.Rmd
trevorcampbell: I think this just makes the sentence more awkward: "you can run X by installing Y" -- installing Y doesn't run X, it just sets things up to run X. Also JupyterLab Desktop is distinct from JupyterLab. I'm going to keep this one as-is I think.

source/setup.Rmd
trevorcampbell: I thought about this, and decided that it's too granular / out of scope to start teaching people how to use the command line

source/setup.Rmd
trevorcampbell: I like this change. I'll give the original and your proposed version a read as I go through the chapter + your comments

source/setup.Rmd
trevorcampbell: I tried to do this originally but struggled to add it without making a mess. I'll try again...no guarantees!

source/setup.Rmd
trevorcampbell: yes!

source/setup.Rmd
trevorcampbell: Yeah, we should add it if it isn't there... this might be related to forward refs from other areas. Anyway will double check

source/setup.Rmd
trevorcampbell: good catch!

source/setup.Rmd
trevorcampbell: That's just a caption error (good catch) -- it's showing it under windows, I just copy/pasted the caption from the ubuntu section and forgot to change the name

source/setup.Rmd
trevorcampbell: I reworded slightly to just put JLab Desktop first

source/setup.Rmd
trevorcampbell: yup, sounds fine to me

source/setup.Rmd
trevorcampbell: Nope, it makes it a mess (it just means that both windows and mac users have to scroll around to find things, instead of just one of them). I'm going to change to your suggested Windows, Mac, Ubuntu order, but keep these in their own sections.

source/setup.Rmd
trevorcampbell: added a remark to open the drop down if necessary

PR #520, opened 2023-08-21
Remove geom_smooth and use the same viz approach as for knn instead

joelostblom:
@trevorcampbell  Here are the corresponding R changes for https://github.com/UBC-DSCI/introduction-to-datascience-python/pull/207

### DIFF
--- a/source/regression2.Rmd
+++ b/source/regression2.Rmd
@@ -54,7 +54,6 @@ By the end of the chapter, readers will be able to do the following:
 * Use R and `tidymodels` to fit a linear regression model on training data.
 * Evaluate the linear regression model on test data.
 * Compare and contrast predictions obtained from K-nearest neighbor regression to those obtained using linear regression from the same data set.
-* In R, overlay predictions from linear regression on a scatter plot of data using `geom_smooth`.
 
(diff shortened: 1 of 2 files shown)


source/regression2.Rmd
```diff
@@ -292,21 +291,35 @@ sale price based off of the predictor of home size? Again, answering this is
...
+
+sacr_preds <- lm_fit |>
+  predict(sqft_prediction_grid) |>
+  bind_cols(sqft_prediction_grid)
+
 lm_plot_final <- ggplot(sacramento_train, aes(x = sqft, y = price)) +
   geom_point(alpha = 0.4) +
+  geom_line(data = sacr_preds, 
+            mapping = aes(x = sqft, y = .pred), 
+            color = "blue") +
```
joelostblom, line 319: I believe it is redundant to re-specify any mappings inside `aes` in the geom that are the same as the mapping inside `aes` in the initial `ggplot()` call. The explicit mention of the `mapping = ` and `data = ` parameters here is also redundant and different from how they are not written out inside `ggplot()`. However, in both these cases, this is what is done in the rest of the book as well it seems, so I left it as is for consistency.

PR #521, opened 2023-08-25
Use more intuitive column name

joelostblom:
@trevorcampbell Ready for review


trevorcampbell:
@joelostblom it looks like some edits were (?) missed -- e.g. lines 864 and 877 in `inference.Rmd` were unchanged, but in the Py version the corresponding code at line 1065 in `inference.md` was changed.

Please go through and make sure the py and R versions are as in-sync as possible (within reason). Let me know when ready

joelostblom:
Thanks! I didn't notice it because it was named just `mean` instead of `sample_mean` originally. I fixed it plus found a couple of more similar instances.

PR #556, opened 2023-11-10
Better section on paths in reading chapter

trevorcampbell:
corresponding to py book improvements


joelostblom (commented):
LGTM from looking at the diff!

PR #562, opened 2023-11-13
Python sync: predictive chapter improvements

trevorcampbell:
The PR aims to mimic https://github.com/UBC-DSCI/introduction-to-datascience-python/pull/314

### DIFF
--- a/source/regression2.Rmd
+++ b/source/regression2.Rmd
@@ -221,11 +221,11 @@ can come back to after we choose our final model. Let's take care of that now.
 library(tidyverse)
 library(tidymodels)
 
-set.seed(1234)
+set.seed(10)
 
 sacramento <- read_csv("data/sacramento.csv")
 
-sacramento_split <- initial_split(sacramento, prop = 0.6, strata = price)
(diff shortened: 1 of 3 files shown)


source/regression2.Rmd
```diff
@@ -349,7 +349,8 @@ obtained from the same problem, shown in Figure \@ref(fig:08-compareRegression).
 
 ```{r 08-compareRegression, echo = FALSE, warning = FALSE, message = FALSE, fig.height = 4.75, fig.width = 10, fig.cap = "Comparison of simple linear regression and KNN regression."}
 set.seed(1234)
-sacr_spec <- nearest_neighbor(weight_func = "rectangular", neighbors = 30) |>
+# neighbors = 28 from regression1 chapter
```
joelostblom, line 352:
```suggestion
# neighbors = 52 from regression1 chapter
```

joelostblom (commented):
I don't know the R code for this as well as sklearn, but at first glance it looks ok. Only comment is that indentation seems to be a mix of 2 , 4 , and 12 spaces, so aligning that with the rest of the book would be nice.

